PR #187, opened 2024-08-07
Grapefruit bu

nathanaelhuffman:
Merge the basic gfruit stuff into main. We have this working  well on bench and I don't think we need to leave this as a long-running branch any longer.


tools/speeker/example.py
```diff
@@ -6,23 +6,43 @@
...
+#     print(f"0x0: {a:#x}")
+#     con.write32(0x60000200, 0x0)
+    #a = con.read32(0x60000000)
+    #print(f"0x0: {a:#x}")
+
+# #con.write32(0x60000500, 0x0)
+#a = con.read32(0x60000200)
+# print(f"0x0: {a:#x}")
+# a = con.read32(0x60000400)
+# print(f"0x0: {a:#x}")
```
Aaron-Hartwig, line 48: Did we want to leave this commented out for reference only, or?

tools/speeker/ruby_power.py
```diff
@@ -0,0 +1,48 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#
+# Copyright 2024 Oxide Computer Company
+
+# Super simple read/write example
```
Aaron-Hartwig, line 7: Copy/pasted from `example.py`

Aaron-Hartwig (approved):

PR #224, opened 2024-10-04
Utilities: Adds a Strobe block

Aaron-Hartwig:
This PR adds a simple strobe block that issues a single-cycle pulse every `TICKS` clk periods. The intention behind it is you could calculate a constant for `TICKS` using the `time_pkg` to get pulses at various frequencies. I reorganized the utilities folder a bit so we don't end up with a bunch of files just laying around in that directory. I'm open for discussion on organization here.

It also adds the `interfaces_pkg` where we can store commonly used interfaces. I put a one byte wide streaming interface in there but didn't go rework any existing code to use it yet.

Finally, there are a few small fixes to the tb/th generation templates.


hdl/ip/vhd/utilities/packages/interfaces_pkg.vhd
```diff
@@ -0,0 +1,27 @@
...
+-- Copyright 2024 Oxide Computer Company
+
+-- This package relies on the VHDL 2019 feature for "interfaces"
+
+library ieee;
+use ieee.std_logic_1164.all;
+
+package interfaces_pkg is
+
+    type data_channel is record
```
nathanaelhuffman, line 14: If we're going to limit this to 8bits wide, I think we'll need a more specific name for this type? The alternative is to really lean into new stuff and use a package generic to adjust the width. I'm using one in the Xilinx flows here: https://github.com/oxidecomputer/quartz/blob/main/hdl/ip/vhd/axi_blocks/axilite_if_pkg.vhd and you can see it instantiated and used here https://github.com/oxidecomputer/quartz/blob/main/hdl/ip/vhd/espi/espi_target_top.vhd

nathanaelhuffman (approved):

hdl/ip/vhd/utilities/packages/interfaces_pkg.vhd
Aaron-Hartwig: Agreed! I think if we were going to limit it to 8 bits wide it would warrant a more specific name, but making the interface generic is a better design anyway. I'll adjust that.

PR #227, opened 2024-10-11
Fix TofinoDebugPortTest

arjenroodselaar:
#210 missed TofinoDebugPortTest. This diff refactors the tests to use the I2CTestParam struct for initialization and reflect the fact that the controller will ack poll.


Aaron-Hartwig (approved):
:shipit:

PR #232, opened 2024-11-12
Add basic stream VCs

Aaron-Hartwig:
This adds basic stream verification components. The basic stream source/sink occupy a middle ground with VUnit's stream master/slave being simpler (just push/pop of data) and VUnit's AXI/Avalon being richer (more framing and control options). These components allow for pushing and popping of data while respecting the source's `valid` and the sink's `ready`.

### DIFF
--- a/hdl/ip/vhd/vunit_components/BUCK
+++ b/hdl/ip/vhd/vunit_components/BUCK
@@ -26,3 +26,16 @@ vunit_sim(
     deps = [":qspi_controller_vc"],
     visibility = ['PUBLIC'],
 )
+
+vhdl_unit(
+    name = "basic_stream",
+    srcs = glob(["basic_stream/*.vhd"]),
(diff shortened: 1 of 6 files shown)


Aaron-Hartwig:
The sims pass, but for convenient they can be ran with:
```
buck2 run root//hdl/ip/vhd/vunit_components:tb_basic_stream
```

nathanaelhuffman (approved):
One minor question I had was just generally whether the avalon streaming blocks provided or axi streaming blocks provided with vunit would work here instead.  Presumably the avalon ones want SOP and EOP and the axi streaming ones want LAST so maybe not a great fit for all our use-cases.

Aaron-Hartwig:
Yeah unfortunately my intended use case is not really a streaming packet, just an interface that can deal with tx/rx streams via a core state machine.

PR #233, opened 2024-11-13
eSPI Link layer re-write

nathanaelhuffman:
Trying to get this a little more product-ready after learning a bunch with the naive approach.  There is really one key change here with a few knock-on effects:
Split link-layer and transaction layer into different clock domains. We want the link layer to be able to decode qspi traffic at 66MHz, which means we need to run some portion of this design at ~3x this so ~200MHz.  The initial implementation just put the whole block at 200 and dealt with clock crossing into the other blocks but this was resulting in some timing challenges as well as a bunch of annoying clock crossing in different places since the eSPI block communicates with UARTs and SPI nor etc. We had also grown the ability to self-generate transactions which resulted in another clock-cross between the axi domain and this domain.

This change splits the link-layer into its own clock domain with cdc fifos for command and response and restructures the design slightly to accommodate that as well as allowing the debug block to more easily integrate at this fifo-boundary but without clock-crossing.

In order to accomplish this, we have to leverage WAIT_STATE responses as there is not enough turn around time between the command byte's crc and the expected response to go through 2 cdc fifos.  The link layer generates WAIT_STATEs as instructed by the slow domain, which looks them up based on the spi-mode and selected clock rate (our CDC delays are fixed so different spi rates and modes need different #'s of WAIT_STATE cycles to achieve the proper delay).

An annoying "feature" of eSPI is that some number of payload bytes in addition to the opcode are sometimes necessary to determine the command size, and thus where the turn-around cycles are.  The link-layer block had to gain more intelligence about the packets (see cmd_sizer.vhd) in order to deal with this since previously we were relying on the transaction layer to provide single-cycle feedback as to where we were in the packet, and this is no longer possible due to the domain split.

The transaction and higher logic is by and large the same as it was, but we had a few bug fixes where the design was relying on additional cycles between response transfers and now with a FIFO there, the response bytes are fed into the FIFO cycle over cycle vs being paced by the serializer.

This also has a number of running fixes discovered while integrating this with the ruby platform that aren't specifically detailed here, but range from wrong uart pinouts on grapefruit to adjustments for the packet parsing logic that we discovered while working through UART integration. We also still don't have a concrete strategy for the IPCC UART stuff which is now pending some questions back to AMD and likely to cause further changes to some of this design.


hdl/ip/vhd/espi/txn_layer/command_processor.vhd
```diff
@@ -198,7 +206,7 @@ begin
                     v.hdr_idx := 0;
                     -- Now we need to decide where we're going
                     -- options are CRC or HEADER based on opcode
-                    case v.cmd_header.opcode.value is
+                    case? v.cmd_header.opcode.value is
```
Aaron-Hartwig, line 209: Given this is a lesser known (and _relatively new_ language feature), maybe leave a note on `case?` matching on don't care bits (`-`) and how the eSPI spec has them built into its opcodes?

Aaron-Hartwig (approved):
A comment I left in a couple places before figuring I'd just leave more generally is that I'd like a block comment at the top of each file with an entity. It can be detailed or be a summary as you see fit, but I think its critical to help orient someone who may be coming into this subsystem cold. The docs and diagram go a long way here, but I think in-context comments go a long way.

PR #235, opened 2024-11-13
Update synchronizers to use resets, fix broken uart tb

nathanaelhuffman:
This was a hold-over from me trying to be clever and not use async resets for some of these blocks.  I'm generally convinced doing this isn't worth the hassle and lack of clarity.  I found at least one bug or unexpected behavior where when the design comes out of reset but on of the lines here was held high we'd miss the handshake.  Adding back in the various domain resets to tacd resolves that issue with no real functionality change but the interface changes slightly to accommodate the additional resets.

Also fixed uart testbench which was missed on a previous change.


Aaron-Hartwig (approved):
I am pro-reset until it is otherwise necessary. As you state, it makes things very straightforward to reason about. Lets keep them around until they get in the way!

PR #236, opened 2024-11-13
Colo gha

### LINKED ISSUE #141: Get CI running somehow with buck2 and things.
We'd like to run all sims all the time, but ideally only re-build images that have changes.  How?

### LINKED ISSUE #153: Run all simulations on every commit
We should not allow commits with broken simulations to reach `main`.

nathanaelhuffman:
Fixes #153
Fixes #141 

This provides an initial stepping stone toward a more fully features CI and release process, where sims run and FPGA builds run for both cobble and buck2 based builds, with some simple filtering so we don't have to rebuild both sides of the world when something on only one of them changed.

There are a couple of minor edits to things to make the sims go, but no real design-functional changes in here, just CI infra.

I'd like to get this merged, so we can make these required checks going forward as they're providing value already.

We'll need to figure out the release process still, which we probably want some level of automation but likely want it to be on-demand vs a release every merge to main I'd think so it's easier to track down changes.  As it is, you can go fishing through the uploaded artifacts for the things you want to submit and do manual submissions as needed, we just don't have a perma-link to the artifacts yet.


Aaron-Hartwig (approved):
:shipit: 🚀

PR #241, opened 2024-11-21
Initial structure for cosmo hotplug fpga

nathanaelhuffman:
initial pcf, a top file that does nothing and a couple of flow improvements to allow building as this is the first real vhdl+yosys design.


Aaron-Hartwig (approved):
Thanks for getting this stubbed out!

PR #248, opened 2024-12-06
I2C: A couple of fixes around SCL stretching is handled

### LINKED ISSUE #246: QsfpModuleController: SCL stretch detection is too naive for our QSFP control use case
The I2C SCL stretch detection is very simple (https://github.com/oxidecomputer/quartz/blob/e5f3844b0188cf2face99533db33fc96a0e6693f/hdl/ip/bsv/I2C/I2CBitController.bsv#L155-L161): When the I2C controller is not driving SCL, wait a small delay and see if someone else is holding it low. The problem in the context of our QSFP controller is that each module's power supply is only enabled when a module is installed. That rail also energizes the pull-up resistors on the module's I2C rail, so the SCL line will go low. We erroneously detect this as a SCL stretch.

### LINKED ISSUE #247: QsfpModuleController: Need to gate SclStretchTimeout errors stick in some cases
Related to #246

https://github.com/oxidecomputer/quartz/blob/e5f3844b0188cf2face99533db33fc96a0e6693f/hdl/projects/sidecar/qsfp_x32/QSFPModule/QsfpModuleController.bsv#L306-L310

Basically due to #246 when you remove a module we will end up with a `SclStretchTimeout` error. The above code is not working as intended given the comment it seems.

Aaron-Hartwig:
This commit fixes a couple of bugs related to SCL stretching. First off, in the `I2CBitController` I adjusted how we sample SCL to detect stretching so we're not errantly sampling. Then in the `I2CCore` I added a mechanism to handle the SCL stretch timeouts when we see them. Before we lacked this and that meant if that was seen, we were wedged. The rest of the changes were to test cases in order to mimic the behavior observed that brought the problems to our attention.

Fixes #246
Fixes #247

### DIFF
--- a/hdl/ip/bsv/I2C/I2CBitController.bsv
+++ b/hdl/ip/bsv/I2C/I2CBitController.bsv
@@ -48,6 +48,9 @@ interface I2CBitController;
     interface Put#(Event) send;
     interface Get#(Event) receive;
 
+    // way of indicating when there is no bus activity
+    method Bool busy();
+
     // clock stretching information sent sideband from the state machine events
     method Bool scl_stretch_seen();
     method Bool scl_stretch_timeout();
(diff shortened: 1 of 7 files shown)


Aaron-Hartwig:
@nathanaelhuffman I actually did not end up faking out the `I2CCore` per our conversation. In the event a module would get removed while it was being talked to we would detect that as a timeout given the bus voltage will collapse. However now with an actual timeout recovery mechanism in place we can operate as expected when a module comes back since all error/timeout information only pertains to the last transaction.

nathanaelhuffman (approved):
lgtm

PR #250, opened 2024-12-10
Add a countdown module and make strobe VHDL2008 compliant

Aaron-Hartwig:
Also shuffle around a few things in the common BUCK file to make it alphabetical order.

### DIFF
--- a/hdl/ip/vhd/common/BUCK
+++ b/hdl/ip/vhd/common/BUCK
@@ -1,5 +1,49 @@
 load("//tools:hdl.bzl", "vhdl_unit")
 
+# Countdown
+
+vhdl_unit(
+    name = "countdown",
+    srcs = glob(["countdown/countdown.vhd"]),
+    deps = [],
+    visibility = ['PUBLIC'],
+)
+
+vhdl_unit(
+    name = "countdown_tb",
+    is_tb = True,
+    srcs = glob(["countdown/sims/*.vhd"]),
+    deps = [":countdown"],
+    visibility = ['PUBLIC'],
+)
+
+# Interfaces - Note these packages require VHDL-2019 support
+
+vhdl_unit(
+    name = "streaming_if_pkg",
+    srcs = glob(["interfaces/streaming_if_pkg.vhd"]),
+    deps = [],
+    standard = "2019",
+    visibility = ['PUBLIC']
+)
+
+# Strobe
+
+vhdl_unit(
+    name = "strobe",
+    srcs = glob(["strobe/strobe.vhd"]),
+    deps = [],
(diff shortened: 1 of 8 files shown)


hdl/ip/vhd/common/BUCK
```diff
@@ -1,38 +1,23 @@
...
-        ":time_pkg", 
-        ":calc_pkg",
-    ],
+    srcs = glob(["countdown/sims/*.vhd"]),
+    deps = [":countdown"],
     visibility = ['PUBLIC'],
 )
 
-# Interfaces
+# Interfaces - Note these packages require VHDL-2019 support
```
nathanaelhuffman, line 20: Does this add value vs having it defined below?

hdl/ip/vhd/common/countdown/countdown.vhd
```diff
@@ -0,0 +1,59 @@
...
+entity countdown is
+    generic (
+        SIZE    : positive
+    );
+    port (
+        clk     : in std_logic;
+        reset   : in std_logic;
+
+        -- target value to count down from
+        count   : in unsigned(SIZE - 1 downto 0);
```
nathanaelhuffman, line 27: I would consider using ieee.numeric_std_unsigned.all above and then just leaving this as std_logic_vector vs unsigned?

My view generally is everything should be considered unsigned unless we explicitly opt into signed behavior and usign numeric_std_unsigned allows us to use std_logic_vector as unsigned without the cast which is nice.

hdl/ip/vhd/common/countdown/sims/countdown_th.vhd
```diff
@@ -0,0 +1,53 @@
...
+        SIZE    : positive
+    );
+end entity;
+
+architecture th of countdown_th is
+
+    signal clk      : std_logic := '0';
+    signal reset    : std_logic := '1';
+
+    signal dut_count    : unsigned(SIZE - 1 downto 0) := (others => '0');
```
nathanaelhuffman, line 28: same comment here, ieee.numeric_std_unsigned.all which allow you to just use std_logic_vector

nathanaelhuffman (approved):
Some minor nits and optional suggestions.

hdl/ip/vhd/common/BUCK
Aaron-Hartwig: Probably not!

hdl/ip/vhd/common/countdown/countdown.vhd
Aaron-Hartwig: I agree with your rationale here and will implement that.

PR #254, opened 2024-12-12
Add glitch filter to common I2C blocks

nathanaelhuffman:
Adding this little block as the other blocks may also want to use it.

### DIFF
--- a/hdl/ip/vhd/i2c/common/i2c_glitch_filter.vhd
+++ b/hdl/ip/vhd/i2c/common/i2c_glitch_filter.vhd
@@ -1,0 +1,114 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2024 Oxide Computer Company
+
+-- Simple i2c glitch filtering for targets  I2C spec
+-- This block incurs a delay of 1 sync cycle and n filter cycles
+-- on both of these lines.
+-- This filter is implemented as a pipeline of n register stages
+-- with the output registers only being toggled when all n register
+-- outputs are the same.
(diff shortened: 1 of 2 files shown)


hdl/ip/vhd/i2c/common/i2c_glitch_filter.vhd
```diff
@@ -0,0 +1,114 @@
...
+--
+-- Copyright 2024 Oxide Computer Company
+
+-- Simple i2c glitch filtering for targets  I2C spec
+-- This block incurs a delay of 1 sync cycle and n filter cycles
+-- on both of these lines.
+-- This filter is implemented as a pipeline of n register stages
+-- with the output registers only being toggled when all n register
+-- outputs are the same.
+-- we also provide some edge detection signals since we already
```
Aaron-Hartwig, line 13: This is a great call

Aaron-Hartwig (approved):
Looks good! At the risk of being too cute, it occurred to me that this block is basically the same pattern twice. We could make a more distilled block that filtered a signal and provided edge detection and then just instantiate it once each for SCL and SDA. Or you can just merge this :)

nathanaelhuffman:
> Looks good! At the risk of being too cute, it occurred to me that this block is basically the same pattern twice. We could make a more distilled block that filtered a signal and provided edge detection and then just instantiate it once each for SCL and SDA. Or you can just merge this :)

It is true. At least for now, I'm inclined to leave it for the following reasons:
- This code is simple and basically all the business in the process fits neatly on one screen so it can be easily reasoned about.
- breaking it down requires a separate entity saves us a few lines of code at the cost of another other file and entity overhead. I didn't count but it would be at least a wash, if not more code.  Should we find reason to need more complicated logic of filtering in the future though i think this would potentially be warranted.  I did think about doing this in a common procedure also which would be called 2x in the process, but I'm not sure it increases the readability after playing around with that idea a bit.

PR #261, opened 2025-01-16
QSFP: Increase module PG timeout, remove internal pull-downs

### LINKED ISSUE #259: QSFP: Increase  module power supply PG timeout
In at least [one board](https://github.com/oxidecomputer/mfg-troubleshooting/issues/58) and likely on [another](https://github.com/oxidecomputer/mfg-troubleshooting/issues/59) the [current power good timeout of 10 ms](https://github.com/oxidecomputer/quartz/tree/0797b55f2498967389554e101a7a404200b1c0e6/hdl/projects/sidecar/qsfp_x32/QSFPModule/QsfpModuleController.bsv#L64) is insufficient. If I try to remember how I arrived at that number a few years ago it was likely because I measured EN->PG on some ports experimentally and then increased that time by 2-3x. Given this has worked on nearly every single port to date, the problematic ports likely represent a nasty corner case. This is exacerbated by the fact that I debounce the power good input for 5ms, meaning we really need to see PG rise within 5ms.

Given the 10ms number was likely somewhat arbitrary to begin with, I propose we double it to 20ms (effectively 15ms with debouncing). This was validated against a problematic board we've seen in manufacturing.

### LINKED ISSUE #260: QSFP: Remove internal pull-downs on module PG signals
A couple years ago (https://github.com/oxidecomputer/quartz/commit/f05b91b2ce3cc9e285fa04bd762edcdc1b8bcf9f) I added internal pulls on inputs from the module. We don't want pull-downs on the PG signal since the power supply holds it low and then releases it to be pulled up by an external pullup. This pullup is relatively weak in our design at 10K. The divider created by our internal pull-down and this external pull-up meant that our PG signal was sitting at a measly ~2.3V rather than 3.3V. This works because the Vih on the ECP5's input pins has a minimum of 2.0V to resolve to a logical 1, but we should fix this.

This change on its own does not resolve the need for an increased timeout explained in https://github.com/oxidecomputer/quartz/issues/259.

Aaron-Hartwig:
Fixes #259
Fixes #260


Aaron-Hartwig:
Current code where we see a couple of errant power good timeout faults on port 2 and 19:
```
aaron@sam ~ $ pfexec humility -p 0483:3754:0039001D4741500920383733 -a build-sidecar-c-image-default.zip hiffy -c Transceivers.get_module_status
humility: attached to 0483:3754:0039001D4741500920383733 via ST-Link V3
Transceivers.get_module_status() => ModuleStatus { power_enable: 0xffffffff, power_good: 0x50009, power_good_timeout: 0x80004, power_good_fault: 0x0, resetl: 0x0, lpmode_txdis: 0x0, modprsl: 0xfff2fff2, intl_rxlosl: 0x50009 }
```

With this adjustment they go away:
```
aaron@sam ~ $ pfexec humility -p 0483:3754:0039001D4741500920383733 -a build-sidecar-c-image-default.zip hiffy -c Transceivers.get_module_status
humility: attached to 0483:3754:0039001D4741500920383733 via ST-Link V3
Transceivers.get_module_status() => ModuleStatus { power_enable: 0xffffffff, power_good: 0xd000d, power_good_timeout: 0x0, power_good_fault: 0x0, resetl: 0x0, lpmode_txdis: 0x0, modprsl: 0xfff2fff2, intl_rxlosl: 0xd000d }
```

nathanaelhuffman (approved):
lgtm!

PR #266, opened 2025-01-24
Fix various deficiencies around type sharing and types with conflicting reset overrides.

### LINKED ISSUE #265: conflicting reset values for registers that share a type behave unexpectedly
given rdl:

```
reg io {
        field {
            desc = "bits for current port";
        } bits[7:0] =  0;
    };

 io ip0 @ 0x00;
 ip0->name = "Ip0"; 
 ip0->desc = "I/O port input state after polarity inversion (if enabled)";
 ip0.bits->sw = r;

 io ioc0 @ 0x18;
 ioc0->name = "ioc0"; 
 ioc0->desc = "Config registers. 0: output, 1: input";
 ioc0.bits->reset = 0xff;
```
We expected the rec_reset function to return 0's for ip0 and 1's for ioc0, but it does not.
There are a couple of challenges here:
1) We told the RDL generator that these are the same type.
2) rec_reset is an overloaded VHDL function, so it only disambiguates by type, so it can't disambiguate between these two types because they're the same (VHDL2019 allows functions to understand subtypes, so that may be a solve in the long-run but we need to generically
continue supporting VHDL2008 also)
3) We get no error messages in this case, so we don't know anything is wrong, stuff just behaves incorrectly.

Possible next steps:
1) Emit a warning or error message when we have the same type with conflicting reset values. We should probably do this no matter what else we do.
2) I think we only have a few possible options here:
  - Force the user to make these two different types. If they're closely related, this means that they'll have to manually create the overloads to do bitwise operations on them: not super ideal
  - Make the reset optional, and make the user implement two different kinds of rec_reset functions.
  - Somehow figure out the various reset options and make different function names that return the same type but different values. This is sort of a hybrid of the first two options.

nathanaelhuffman:
We had various deficiencies around sharing and overriding reset values on registers that share types. This provides the groundwork for better functioning here and provides error messages when the user asks for something we don't support. We also added new reset_0s and reset_1s functions to all register types so long as they don't have enumerated types. This provides some additional easy knobs to get alternate reset values for your register types.

This isn't at totally generic, full solve but provides some additional tools like the reset_1s, reset_0s functions and at least error messages when things are done that aren't supported currently vs silently doing the unexpected/wrong thing.  We will also elide generating the rec_reset function if a reset value was not set for a given register type.

Fixes #265

### DIFF
--- a/tools/site_cobble/rdl_pkg/listeners.py
+++ b/tools/site_cobble/rdl_pkg/listeners.py
@@ -78,6 +78,7 @@ class BaseListener(RDLListener):
     def __init__(self):
         self.prefix_stack = []
         self.known_types = []
+        self.reset_by_known_type = {}  # A dictionary for looking up reset values by their known-type names
         self.cur_reg = None
         self.registers = []
 
@@ -104,11 +105,6 @@ class BaseListener(RDLListener):
         self.prefix_stack.pop()
 
     def enter_Reg(self, node: RegNode) -> None:
-        # print(f"orig type name: {node.orig_type_name}")
-        # print(f"new type name: {node.type_name}")
-        # print(f"segment: {node.get_path_segment()}")
-        # print(f"Enter reg: {node.inst_name}")
-        # print(f"stack: {self.prefix_stack}")
         # 2 cases here:
         # node.orig_type_name is None, use node.type_name
         #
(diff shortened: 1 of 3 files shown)


nathanaelhuffman:
I need to sort out the bug that's failing ci

tools/site_cobble/rdl_pkg/listeners.py
```diff
@@ -78,6 +78,7 @@ class BaseListener(RDLListener):
     def __init__(self):
         self.prefix_stack = []
         self.known_types = []
+        self.reset_byte_known_type = {}
```
Aaron-Hartwig, line 81: Regarding the name here, there's nothing saying the register (and its reset value) would be a single byte. Maybe just `reset_known_types` instead? I also made it plural since this is dictionary that will hold a number of values.

tools/site_cobble/rdl_pkg/listeners.py
```diff
@@ -136,6 +137,31 @@ def exit_Reg(self, node):
         # print(f"Exit reg: {node.inst_name}")
```
Aaron-Hartwig, line 137: Could sweep this up.

tools/site_cobble/rdl_pkg/models.py
```diff
@@ -111,27 +111,67 @@ def __init__(self, **kwargs):
...
+                rst_prop != field.reset_1s):
+                    return False
+        return True
+    
+    @property
+    def reset_is_all_0s(self) -> bool:
+        """
+        True if all defined, non-reserved fields have a reset value of all 0s
+        """
+        return self.elaborated_reset == 0
```
Aaron-Hartwig, line 156: It jumps out to me how `reset_is_all_1s` looks so different than `reset_is_all_0s` when they essentially answer the same thing. If I'm understanding correctly, the 1s just looks field by field to check for 1s, while `elaborated_reset` will hand back the reset for the whole register. We can rely on `elaborated_reset` for the 0s case because we define `ReservedFields` to be 0 always. Am I following correctly?

tools/site_cobble/rdl_pkg/templates/regpkg_vhdl.jinja2
```diff
@@ -239,6 +253,26 @@ package body {{module_name}} is
         {% endfor %}
         return ret_rec;
     end rec_reset;
+    {% endif %}
+    {# We can only generate reset1's and reset0's for non-enumerated types #}
+    {% if not register.has_encoded_fields %}
+    function reset_1s return {{reg_type_name}} is
+        variable ret_rec : {{reg_type_name}};
+      begin
```
Aaron-Hartwig, line 261:
```suggestion
    function reset_1s return {{reg_type_name}} is
        variable ret_rec : {{reg_type_name}};
    begin
```

tools/site_cobble/rdl_pkg/templates/regpkg_vhdl.jinja2
```diff
@@ -239,6 +253,26 @@ package body {{module_name}} is
...
+        {% for field in register.packed_fields %}
+        ret_rec.{{field.name}} := {{field.vhdl_reset_1s}};
+        {% endfor %}
+        return ret_rec;
+    end reset_1s;
+    function reset_0s return {{reg_type_name}} is
+        variable ret_rec : {{reg_type_name}};
+      begin
```
Aaron-Hartwig, line 269:
```suggestion
    function reset_0s return {{reg_type_name}} is
        variable ret_rec : {{reg_type_name}};
    begin
```

Aaron-Hartwig (approved):
One naming suggestion and a question, otherwise this looks good!

tools/site_cobble/rdl_pkg/models.py
nathanaelhuffman: Yes, the ReservedFields don't contribute to the reset value which is otherwise shifted and bit-wise OR'd in so any '1's in the defined bits make this a non-0 but if it's 0, everything defined was zero.

The reset 1s is more complicated b/c we have to actively avoid looking at the ReservedField contribution since they aren't showing as 1s.

tools/site_cobble/rdl_pkg/listeners.py
nathanaelhuffman: i think this is just a typo, it's a dictionary so I think I intended it to be reset_*by*_known_type since the known types are the dict keys. Will fix up.

PR #269, opened 2025-01-28
Add PCA9506ish i/o expander logic block, deal with multi-targets and fix mux features

nathanaelhuffman:
This block works as an emulated PCA9506 I/O expander.
Tri-state logic control left to the user at the top.  Pretty good test coverage on this one, and it leverages the common i2c target phy block like the mux also did.

This also adds a few more outputs (so backwards compatible changes) to the target phy to expose more info like start and stop which allows the upstream state machines to recover more gracefully.

This also fleshes out the missing "mixed-mode" i2c packets in the simulation controller where the controller can issue write-repeated-start-read type i2c transactions.

I added a multi-target consolidator, and multi-target test  cases to both the I/O expander and the mux. Doing this uncovered a few bugs related to bus sharing, which have now been addressed and all the sims and the new and old test cases are running well.


hdl/ip/vhd/i2c/muxes/PCA9545ish/pca9545ish_top.vhd
```diff
@@ -40,6 +40,7 @@ entity pca9545ish_top is
 
         -- Signal to reset the mux-state out of band if desired.
         mux_reset: in std_logic;
+        other_mux_selected: in std_logic := '0';
```
Aaron-Hartwig, line 43: Is there a reason this port is getting a default value? I'm assuming it would be for cases when you'd only be using one instance? I am generally against defaults on parameters and ports regardless.

hdl/ip/vhd/i2c/muxes/PCA9545ish/pca9545ish_function.vhd
```diff
@@ -70,7 +77,7 @@ architecture rtl of pca9545ish_function is
 begin
 
     inst_ready <= '1';  -- never block writes
-    resp_valid <= '1'; -- never block reads
+    resp_valid <= is_our_transaction; -- never block reads
```
Aaron-Hartwig, line 80: The comment here should get updated now that reads can block.

hdl/ip/vhd/i2c/io_expanders/PCA9506ish/pca9506_pkg.vhd
```diff
@@ -0,0 +1,77 @@
...
+
+    type pca9506_pin_t is array (0 to 4) of std_logic_vector(7 downto 0);
+
+    function get_irq_pend(
+        cur_reg: io_type;
+        reg_at_last_read: io_type;
+        reg_mask: io_type
+    ) return std_logic;
+
+    function category_wrapping_increment(
```
Aaron-Hartwig, line 33: Maybe add a comment as to what a "category" is? I flipped through the datasheet before reviewing this so I understand what you're going for here, but a little explanation may help someone who lands in this cold.

Aaron-Hartwig (approved):
LGTM

hdl/ip/vhd/i2c/muxes/PCA9545ish/pca9545ish_top.vhd
nathanaelhuffman: I knew I was going to get this comment when I wrote this :D.  This is for when you have only one instance, or I suppose if you didn't care if multiple muxes were enabled at the same time.

Philosophically, I think I see more shades of grey here on defaults, and I tend to preferring the most simple interface for the common use-case (which would allow one to totally drop the port from the port map), especially when the stakes aren't that high, or where you expect one to almost always want a specific option and only bail to the other as a rare exception, but for this use-case it doesn't matter that much so I'll save my ammunition for a usecase I care more strongly about should it exist in the future.

hdl/ip/vhd/i2c/muxes/PCA9545ish/pca9545ish_function.vhd
nathanaelhuffman: That's a good thought, though it actually points out the fact that this isn't 100% accurate and "hides" what is actually going on here.  With all of this stuff, the clock waits for no-one so backpressure doesn't really happen here, but this was done when I was chasing a bug where the phy was  responding when it wasn't "our transaction" because response data was valid. 

I later realized that this fix masked the real bug, which was that the phy layer was accepting responses even when it wasn't selected, rather than just rolling back to IDLE and waiting for the next start.  I discovered that with the dual i/o expanders and resolved it properly but this stayed as I forgot about it.  I'll flip this back to always enabled since we'll only ever ask for responses when it is our transaction now and there's no backpressure.

hdl/ip/vhd/i2c/muxes/PCA9545ish/pca9545ish_top.vhd
nathanaelhuffman: I removed the default and changed the port name to be a bit more clear since it could also be used to disable muxing when the downstream devices are powered off or something like that.

PR #270, opened 2025-02-03
Add shift_in_at_0 and shift_in_at_high functions 

nathanaelhuffman:
This stuff is pretty mechanical so I didn't write out a super detailed tb but did add cases that cover slv, unsigned and their reverse-range forms.

### DIFF
--- a/hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
+++ b/hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
@@ -2,7 +2,7 @@
 -- License, v. 2.0. If a copy of the MPL was not distributed with this
 -- file, You can obtain one at https://mozilla.org/MPL/2.0/.
 --
--- Copyright 2024 Oxide Computer Company
+-- Copyright 2025 Oxide Computer Company
 
 library ieee;
 use ieee.std_logic_1164.all;
@@ -27,6 +27,11 @@ begin
     bench: process
         variable calc_result : unsigned(31 downto 0);
         variable tmp_flip    : std_logic_vector(7 downto 0);
+        variable tmp_slv_right_lsb     : std_logic_vector(7 downto 0);
+        variable tmp_slv_left_lsb : std_logic_vector(0 to 7);
+        variable tmp_unsigned_right_lsb : unsigned(7 downto 0);
+        variable tmp_unsigned_left_lsb : unsigned(0 to 7);
+
     begin
         -- Always the first thing in the process, set up things for the VUnit test runner
         test_runner_setup(runner, runner_cfg);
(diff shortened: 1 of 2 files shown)


hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
```diff
@@ -46,9 +51,49 @@ begin
                 check_equal(calc_result, 100_000, "Nope");
                 calc_result := calc_ns(desired_ns => 10, clk_period_ns => 10, return_size => 32);
                 check_equal(calc_result, 1, "Nope");
-            elsif run("transform_pkg_test") then
+            elsif run("transform_pkg_reverse_test") then
                 tmp_flip := reverse(X"0F");
                 check_equal(tmp_flip, std_logic_vector'(X"F0"), "Nope");
+            elsif run("transform_pkg_shift_in_at_0_test") then
+                tmp_slv := x"AA";
+                tmp_slv_rev := x"44";
```
Aaron-Hartwig, line 59:
```suggestion
                tmp_slv_rev := reverse(tmp_slv);
```

hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
```diff
@@ -46,9 +51,49 @@ begin
...
+                tmp_unsigned := x"AA";
+                tmp_unsigned_rev := x"44";
```
Aaron-Hartwig, line 61:
```suggestion
                tmp_unsigned_rev := reverse(tmp_unsigned);
```

hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
```diff
@@ -46,9 +51,49 @@ begin
...
+
+                -- slv and unsigned, non-reversed
+                tmp_slv := shift_in_at_0(tmp_slv, '1');
+                check_equal(tmp_slv, std_logic_vector'(x"55"), "Incorrect shift in at 0, slv");
```
Aaron-Hartwig, line 65: In a similar vein to making use of our extant `reverse` function, should we make use of `shift_left` and `shift_right` in these tests since shifting by 1 is the function we're trying to make more ergonomic?
```suggestion
                check_equal(tmp_slv, shift_left(tmp_slv, 1), "Incorrect shift in at 0, slv");
```

Aaron-Hartwig (approved):
Just some notes about where I think we could use functions rather than hardcoded values but they can also be ignored given they don't really make much real difference in this context (no LOC reduction, the tests already run, etc).

hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
nathanaelhuffman: interestingly this isn't actually what is going on (you'll note that reversing x"AA" does *not* get you x"44").  These variables where meant to be a standard range slv (x donwto 0) and a reverse range slv (0 to x) but not actually the bitwise reversal of each other, I *could* have written the test that way, but I chose values that are easy to understand when testing and not necessarily reversable for that same purpose.

hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
nathanaelhuffman: so this doesn't actually work in that test case, and is *why* we wanted these functions.

This is my code, shifting a 1 in at 0 (a left shift + insert of x"AA" resulting in x"55")
```
tmp_slv := x"AA";
 tmp_slv := shift_in_at_0(tmp_slv, '1');
 check_equal(tmp_slv, std_logic_vector'(x"55").....);
```

Your suggestion would read like this, but notably misses the shifting in of a 1 at bit 0, shift_left takes the vector to shift and the *shift_count* not the new input bit.  This also doesn't necessarily work with slvs without numeric_std_unsigned, but assuming an unsigned value here, this will shift in a 0 at the lsb with no control of that.
```
tmp_slv := x"AA";
 check_equal(tmp_slv,  shift_left(tmp_slv, 1)....);
```
This results in x"54" not x"55" because we shifted in a 0.

hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
nathanaelhuffman: see above, this is not equivalent.

hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
Aaron-Hartwig: I agree with your assessment and I apologize that my hasty comment did not capture the spirit of what I intended by the comment. I understand why the suggested code is not equivalent. 

The intention with these functions is to make something that is nicer than:
```
var := shift_left(vec, 1)(vec'high-1 downto vec'low) & '1'; -- or '0'
```
by making
```
var := shift_in_at_0(vec, '1'); -- or '0'
```
So my intention in my comment was to make the point that we should assert equivalence between the thing we wrote and the thing it intends to replace, rather than simply shortcut to the answer because we know it.

hdl/ip/vhd/common/utils/sims/utilities_tb.vhd
nathanaelhuffman: I re-wrote all of these tests using a more simple testcase where we just shift 1 bit into 0, whether its high or low, maybe that's more clear?

Aaron-Hartwig: Ah, doh. I was just glancing at the naming and wrongly assumed that the `$_rev` was the reverse of `$`. Given its not really a reversal, should these be named something that describes their endian difference?

nathanaelhuffman: ok that's done

Aaron-Hartwig (approved):
Ship it!

PR #279, opened 2025-02-07
I2C controller: add ability to abort transactions

Aaron-Hartwig:
A feature we need for the SPD mux testing is the ability for our controller to abort an in-progress transaction. This commit adds an `abort` port to the `i2c_ctrl_txn_layer` entity which will react to a pulse of that signal. In practice this means if the controller is writing it will send a STOP instead of the next bit. When reading, the controller will finish reading the byte, NACK, and then send a STOP.

### DIFF
--- a/hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
+++ b/hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
@@ -266,8 +266,13 @@ begin
             v.transition_sda_cntr_en := '0';
         end if;
 
-        -- Latch if a stop is requested to handle the case when it may need to happen before the end
-        -- of the transaction.
+        -- Set if a stop is requested to handle the case when it may need to happen before the end
+        -- of the transaction. This is then cleared as transition out of the STOP_SETUP state.
+        -- The combinatorial value of this signal is read elsewhwere in the state machine in an
+        -- effort to accommodate making the stop happen as quickly as possible. The intention is
+        -- to transition to a STOP during a normal transition point for I2C which is not necessarily
+        -- any cycle of the FPGA clk. At slower operating modes this does not matter much, but at
+        -- FAST_MODE_PLUS (1MHz) it starts to matter more.
         if sm_reg.stop_requested = '0' then
             v.stop_requested := '1' when tx_stop = '1' and txn_next_valid = '1' else '0';
         end if;
@@ -398,7 +403,7 @@ begin
                         -- at the next transition point release the bus
                         v.sda_oe        := '0';
                         v.ack_sending   := '0';
-                        v.state         := STOP_SDA when sm_reg.stop_requested else HANDLE_NEXT;
+                        v.state         := STOP_SDA when v.stop_requested else HANDLE_NEXT;
                     end if;
                 end if;
 
@@ -426,9 +431,8 @@ begin
 
         end case;
 
-        -- next state logic
         v.ready := '1' when v.state = IDLE or
-                            (v.state = HANDLE_NEXT and sm_reg.stop_requested = '0')
+                            (v.state = HANDLE_NEXT and v.stop_requested = '0')
                     else '0';
 
         sm_reg_next <= v;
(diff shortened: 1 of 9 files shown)


hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
```diff
@@ -311,12 +328,14 @@ begin
                 end if;
 
             when HANDLE_NEXT =>
-                if txn_next_valid then
+                if v.stop_requested then
```
nathanaelhuffman, line 331: One challenge with the variables and registers is that when you compare a variable, you're building a combinatorial path between that and your checking logic.  If you need access to this variable *the same* cycle it was set, this is correct and acceptable, but if you intend to look at it one or more clock later, you're better off using the sm_reg. value as you'll get it as an output from the register.

You have a mix of these in this block, and while it may be what you intended, it is unclear why you'd pick the variable value or the registered value in some cases.  Suggest evaluating and adding clarifying comments as needed.  I'll point out some  below.

The general pattern should be use the register value unless you need to same-cycle value. I didn't fully analyze this to decide which was needed here.
I find that auto-complete has a harder time when the register isn't named "r" and often substitutes v when I'm writing in this style which can lead to typos.

hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
```diff
@@ -370,13 +391,14 @@ begin
...
-                        v.sda_oe        := tx_ack;
+                        v.sda_oe        := '1' when (tx_ack = '1' and v.stop_requested = '0')
+                                            else '0';
                         v.ack_sending   := '1';
                     else
                         -- at the next transition point release the bus
                         v.sda_oe        := '0';
                         v.ack_sending   := '0';
-                        v.state         := HANDLE_NEXT;
+                        v.state         := STOP_SDA when sm_reg.stop_requested else HANDLE_NEXT;
```
nathanaelhuffman, line 401: did you mean sm_reg.stop_requested or v.stop_requested and since this one is different than the others, why?

hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
```diff
@@ -405,7 +427,9 @@ begin
         end case;
 
         -- next state logic
```
nathanaelhuffman, line 429: This whole process is "next state logic" not just this little bit

hdl/ip/vhd/i2c/controller/txn_layer/i2c_ctrl_txn_layer.vhd
```diff
@@ -223,6 +233,12 @@ begin
                 end if;
         end case;
 
+        -- if a transaction is in progress and abort is asserted we should immediate STOP
+        if abort = '1' and v.state /= IDLE and v.state /= STOP and v.state /= WAIT_STOP then
```
nathanaelhuffman, line 237: do you need to check the next state you're transitioning to (v.state) or mean to check your current state (sm_reg.state) here?

nathanaelhuffman (approved):
A couple of comments on whether you're intending to us current registered state vs forwarded comb next state logic in some conditionals, otherwise looks fine.

hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
Aaron-Hartwig: I'll add a clarifying comments up where I drive the signal. My intention _is_ to check it on the combinatorial path (the two cases where I use registered version don't make sense to me anymore with a fresh brain). I understand the timing pressure implications but I want to attempt to react to the abort as quickly as possible since it is not as simple as "we could just do it next cycle" because we cannot always just send a STOP. This is an optimization to catch the situation where we'd otherwise be transitioning to a new state and would not evaluate our ability to send a STOP again for some amount of time. It really doesn't make a lot of sense outside of the FastMode+ use case since there our usual FPGA clk cycle is a meaningful amount of time, so if it is causing problems I'll revisit this later.

hdl/ip/vhd/i2c/controller/txn_layer/i2c_ctrl_txn_layer.vhd
Aaron-Hartwig: This was a brain fart on my part. Thank you for catching it!

PR #282, opened 2025-02-10
QSFP Controller: make power control and readback separate

### LINKED ISSUE #280: QSFP: want HW pin state for if power is _actually_ enabled
Currently the power_enable bit is a SW writable bit which means power should be applied to a port. In practice, we only apply that power once a module's presence has been detected. However, SW has no visibility into this which means we get confusing reports like:
```
root@oxz_switch0:~# xcvradm -i sidecar0 status
Port Status
   0 ENABLED | RESET | INTERRUPT
...
```
Where `ENABLED` is set (because it defaults to `1`), effectively representing the "software enabled" state. This should represent the hardware state of the pin.

Aaron-Hartwig:
The current state of upstack software understanding module power control is confusing. The actual hardware power control had an additional level of gating the FPGA that enforced a module being present before turning power on. However, when software comes by to read the `POWER_ENx` registers it sees the modules as being enabled. This is confusing!

This PR splits out the software control of the power enable into a separate `SW_POWER_ENx` registers from the true pin-level hardware state (post-gating) shown in the `POWER_ENx` registers.

The software impact of this changes are minimal. When fetching state the `POWER_ENx` register will represent the what a user would expect. When modifying if a power _should_ be enabled or not, the `SW_POWER_ENx` registers will need to be hit instead.

fixes #280

### DIFF
--- a/hdl/projects/sidecar/qsfp_x32/QSFPModule/QsfpModuleController.bsv
+++ b/hdl/projects/sidecar/qsfp_x32/QSFPModule/QsfpModuleController.bsv
@@ -97,7 +97,7 @@ interface QsfpModuleController;
     method Action lpmode(Bit#(1) val);
 
     // Software controlled hot swap controller power enable
-    method Action power_en(Bit#(1) val);
+    method Action sw_power_en(Bit#(1) val);
 
     // Power fault state from the controller
     method Bool pg;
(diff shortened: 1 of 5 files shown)


nathanaelhuffman (approved):
LGTM.

Issue #245, opened 2024-12-04
Want a SPI -> AXI interface for the front hotplug fpga controller


nathanaelhuffman:
We're likely going to want to be able to peek/poke this from hubris and should leverage the protocol already used by sidecar and gimlet, but with an axi interface and decoder.

PR #283, opened 2025-02-11
Spi axi controller

nathanaelhuffman:
In the same spirit of the i2c block, this has a spi target phy and then a state machine that does something useful with the data, in this case an axi-lite controller. Fixes https://github.com/oxidecomputer/quartz/issues/245

bit-set and bit-clear functionality was implemented to match feature parity with the existing bsv module, but is done using an automated read-modify-write pattern since AXI-lite interfaces do not support atomic bitset/bitclear operations.

Also note that axi-lite supports a minimum address size of 32bits, this is broken down into individual 8bit transactions using wstrobes on the write side, so peripherals being accessed by this with write access need to be adjusted to respect the wstrobe byte enables (https://github.com/oxidecomputer/quartz/pull/278 makes this more palatable to users).
Reads are just re-read each time and the bits not requested aren't cached. This is the simplest solution, but fancier things could be done so long as register side-effects are considered.

I also noticed some typos in the comments of other blocks and a simple mis-math with signed numbers that was discovered by inspection, not something we're likely to run into but that was resolved there also.

I noticed asserts were not stripped in all cases when using the ghdl+yosys toolchain, this adds --no-formal to the ghdl plugin to strip the asserts as suggested here: https://github.com/ghdl/ghdl-yosys-plugin/issues/210#issuecomment-2652767427


hdl/ip/vhd/spi/axi_controller/spi_axi_sms.vhd
```diff
@@ -0,0 +1,356 @@
...
+                end if;
+            when WAIT_READ_RESP =>
+                -- wait for axi response
+                if axi_r.state = IDLE then
+                    v.data := axi_r.rdata;
+                    v.to_spi_valid := '1';
+                end if;
+                if to_spi_valid = '1' and to_spi_ready = '1' then
+                    v.to_spi_valid := '0';
+                    v.req_addr := v.req_addr + 1;  -- increment and go again
```
Aaron-Hartwig, line 183: Can be a future work item, but I think we will want to support a `{READ,WRITE}_NO_ADDR_INCR` operation to expose control over if the address should be incremented or not. We support (and use!) this in our BSV SPI core.

hdl/ip/vhd/spi/axi_controller/spi_axi_sms.vhd
```diff
@@ -0,0 +1,356 @@
...
+                -- get data from spi, move to wait write response
+                if from_spi_valid and from_spi_ready then
+                    v.data := from_spi_data;
+                    v.from_spi_ready := '0';
+                    v.state := ISSUE_WRITE;
+                end if;
+            when ISSUE_WRITE =>
+                if axi_r.state = IDLE then
+                    v.from_spi_ready := '1';
+                    v.req_addr := v.req_addr + 1;  -- increment and go again
```
Aaron-Hartwig, line 204: See above regarding ability to control address incrementing at the transaction level.

hdl/ip/vhd/spi/axi_controller/spi_axi_pkg.vhd
```diff
@@ -0,0 +1,79 @@
...
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+package spi_axi_pkg is
+
+    constant spi_opcode_write : std_logic_vector(3 downto 0) := "0000";
+    constant spi_opcode_read : std_logic_vector(3 downto 0) := "0001";
+    constant spi_opcode_bit_set : std_logic_vector(3 downto 0) := "0010";
+    constant spi_opcode_bit_clr : std_logic_vector(3 downto 0) := "0011";
```
Aaron-Hartwig, line 16: I've got a comment below about adding support for non-address incrementing reads and writes. That can be future work.

`b0101` for WRITE_NO_ADDR_INCR (https://github.com/oxidecomputer/hubris/blob/41792abce4acdc6a8e822647520d894f3bf23bae/drv/fpga-api/src/lib.rs#L145)
`b0110` for READ_NO_ADDR_INCR (https://github.com/oxidecomputer/hubris/blob/41792abce4acdc6a8e822647520d894f3bf23bae/drv/fpga-api/src/lib.rs#L169)

hdl/ip/vhd/spi/spi_target_phy/spi_target_phy.vhd
```diff
@@ -0,0 +1,205 @@
...
+            -- upstream needs to beat our need for data by at least
+            -- one clk (fast, not sclk) cycle.  This should be no problem
+            if csn_syncd = '1' then
+                tx_buffer.empty <= '1';
+            elsif tx_valid = '1' and tx_ready = '1' then
+                tx_buffer.data <= tx_data;
+                tx_buffer.empty <= '0';
+            end if;
+
+            --
```
Aaron-Hartwig, line 178: A whisper in the void

hdl/ip/vhd/spi/axi_controller/sims/spi_axi_tb.vhd
```diff
@@ -0,0 +1,206 @@
...
+    th: entity work.spi_axi_th;
+
+    bench: process
+        alias reset is << signal th.reset : std_logic >>;
+        alias csn is << signal th.csn : std_logic >>;
+        variable buf           : buffer_t;
+        variable buf2          : buffer_t;
+        variable data          : std_logic_vector(7 downto 0) := (others => '0');
+        variable expected_data : std_logic_vector(7 downto 0) := (others => '0');
+        variable address       : std_logic_vector(15 downto 0) := (others => '0');
```
Aaron-Hartwig, line 43: Is it worth an asymmetrical address here (e.g. `xAA55`) to help catch any byte swapping issues? I don't see it getting set to anything else elsewhere.

Aaron-Hartwig (approved):
This is great! I mostly just had a couple questions, but one thing I did want to capture is that I think we want our SPI/AXI stuff to support not automatically incrementing addresses. I think that can definitely be outside of this PR and I'm happy to write the issue for future-us.

None
```diff
@@ -0,0 +1,11309 @@
+{
```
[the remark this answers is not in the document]
nathanaelhuffman: jj!  Yeah I'll remove this.  an artifact of jj and me doing too many things in too many different branches.

hdl/ip/vhd/spi/axi_controller/docs/README.adoc
```diff
@@ -0,0 +1,15 @@
...
+|Opcode| Instruction Name | Description| Notes
+| 0x0 | Write byte(s)      | Write one or more contiguous bytes | 
+| 0x1|  Read byte(s)      | Read one or more contiguous bytes|
+| 0x2| Bit set | hardware does a bit-wise OR with data and current register state | new_reg = old_reg \|\| data
+| 0x3| Bit clr | hardware does a bit-wise clear with data and current register state |new_reg = old_reg && !data
+|===
+
+Note that for the bit-set and bit-clear spi instructions, the AXI controller internally does a read-modify-write operation, resulting
+in a read of the current register state, a bitwise operation with the data, and a write back of the new register state.  These are
+not good operations to use if the registers you're modifiying have read-side effects.
```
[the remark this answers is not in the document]
nathanaelhuffman: Yeah I'm medium on this when I wrote it but now realizing that all the avalon interfaces are 32bit due to axi-lite limitations, I actually think we may want to allow this and test it. It's kind of funky in that you'd have to provide the byte-wise values in consecutive commands, but I think I'll just write tests for this and allow it actually even if it mostly won't be used.

hdl/ip/vhd/spi/axi_controller/spi_axi_controller_top.vhd
```diff
@@ -0,0 +1,123 @@
...
+        -- spi pins
+        -- We're running these through meta sync so there's a rate limit
+        -- here, related to your clock frequency
+        csn  : in std_logic;
+        sclk : in std_logic;
+        copi : in std_logic;
+        cipo : out std_logic;
+
+        -- Would love to use 2019 record views here, but we're stuck with
+        -- 2008 compat for the ice40 and open toolchain flows
```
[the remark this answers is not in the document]
nathanaelhuffman: certainly possible.  It is unclear what designs we might want a SPI-AXI master that *also* could use VHDL-2019 in their toolchain, but should such a design present itself, I'm 100% game for making wrappers.  I think that's the general disposition for shared code until such a time as all of our tools support 2019.  For stuff like this where were guaranteed to need it in a 2008 world I'm writing it this way but will be the first to support wrappers should it be needed in a 2019-eligible design.

hdl/ip/vhd/spi/axi_controller/spi_axi_sms.vhd
```diff
@@ -0,0 +1,356 @@
...
+-- See docs for details on the SPI protocol
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+use ieee.numeric_std_unsigned.all;
+
+use work.spi_axi_pkg.all;
+
+entity spi_axi_sms is
```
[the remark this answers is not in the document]
nathanaelhuffman: well, our toolchains don't support imessage or rcs....  do you think this should change?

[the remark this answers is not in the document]
Aaron-Hartwig: I blame Apple!

I think something like `spi_to_axi` or something may be more appropriate. The "spi axi" name doesn't really tell you what is happening on its face.

hdl/ip/vhd/spi/axi_controller/spi_axi_pkg.vhd
nathanaelhuffman: I have context now and I think it's not that hard to patch this in, will add and add test coverage.

hdl/ip/vhd/spi/axi_controller/spi_axi_sms.vhd
nathanaelhuffman: will implement the non-incr opcodes also

hdl/ip/vhd/spi/axi_controller/spi_axi_sms.vhd
nathanaelhuffman: will implement + test

hdl/ip/vhd/spi/axi_controller/spi_axi_sms.vhd
```diff
@@ -0,0 +1,356 @@
...
+                end if;
+
+            when ADDR1 =>
+                if from_spi_valid and from_spi_ready then
+                    v.req_addr(15 downto 8) := from_spi_data;
+                    v.state := ADDR2;
+                end if;
+            when ADDR2 =>
+                if from_spi_valid and from_spi_ready then
+                    v.req_addr(15 downto 8) := from_spi_data;
```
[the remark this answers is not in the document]
nathanaelhuffman: Nope, just is actually just a bug. this should be 7 downto 0.  good catch. This also points out a deficiency in my testing approach where I've always started from 0, though I'm not sure malloc'ing enough buffer space to cover 16bit addressing makes sense here.  I may try to start at a non-zero address though which should work.

hdl/ip/vhd/spi/axi_controller/sims/spi_axi_tb.vhd
nathanaelhuffman: I added a new test-case run("non-zero-addr-read") that does a read at 1.  This ensures that we're not always reading from 0 where the top and bottom addresses are both 0.  I confirmed that with my original double 15..0 bug, this test case trips and it passes with my fix.

hdl/ip/vhd/spi/axi_controller/docs/README.adoc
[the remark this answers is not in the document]
nathanaelhuffman: I added test cases, for this. It works out of the box, but is a bit of a strange pattern given that you have to provide each of the bytes for the bit-set or bit-clear, but the hw does work as the test cases demonstrate.

hdl/ip/vhd/spi/axi_controller/spi_axi_pkg.vhd
```diff
@@ -33,19 +36,27 @@ package body spi_axi_pkg is
...
         -- known opcodes are 0-3
-        return unsigned(opcode) < 4;
+        return unsigned(opcode) < 4 or 
+            opcode = spi_opcode_write_no_addr_incr or
+            opcode = spi_opcode_read_no_addr_incr;
     end function;
 
     function is_read_kind_opcode(opcode : std_logic_vector(3 downto 0)) return boolean is
     begin
         -- only read opcode we have
```
Aaron-Hartwig, line 46: maybe update this comment to be similar to what you put in down at L57

hdl/ip/vhd/spi/axi_controller/docs/README.adoc
```diff
@@ -1,15 +1,15 @@
...
 | 0x1|  Read byte(s)      | Read one or more contiguous bytes|
 | 0x2| Bit set | hardware does a bit-wise OR with data and current register state | new_reg = old_reg \|\| data
 | 0x3| Bit clr | hardware does a bit-wise clear with data and current register state |new_reg = old_reg && !data
-|===
-
-Note that for the bit-set and bit-clear spi instructions, the AXI controller internally does a read-modify-write operation, resulting
-in a read of the current register state, a bitwise operation with the data, and a write back of the new register state.  These are
-not good operations to use if the registers you're modifiying have read-side effects.
\ No newline at end of file
+| 0x5| Write bytes(s) no increment | Write one address location one or more times (internal byte address does not increment)
```
Aaron-Hartwig, line 14: Should also add a line for `0x6 Read byte(s) no increment`.

Aaron-Hartwig (approved):
A couple of doc and comment related things but otherwise feel free to merge this!

PR #285, opened 2025-02-13
Improve logging for the yosys toolchain

nathanaelhuffman:
The yosys toolchain was mostly slapped together as proof of concept. After using it in anger for a bit I've found some missing functionality around the logs we're exposing so this exposes more of the correct logs, like we'd like to get warnings from the ghdl synthesizer, the yosys log and the next pnr log with area and timing.

The ghdl thing is a bit strange in that it logs warnings to stderr when run as a plugin in yosys but the warnings don't make it into the log file, probably due to how yosys is logging things. I've captured the stderr there and written it to a file, and printed it to stderr so that when we get a non-zero exit you see the errors in the terminal.

### DIFF
--- a/tools/yosys.bzl
+++ b/tools/yosys.bzl
@@ -73,13 +73,19 @@ def yosys_vhdl_synth(ctx):
     cmd.add(yosys_gen)
     cmd.add("--input", in_json_file)
     cmd.add("--output", yosys_py.as_output())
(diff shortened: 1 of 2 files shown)


Aaron-Hartwig (approved):
Thanks for further fleshing this flow out!

PR #287, opened 2025-02-14
Axi 2k8 work and cleanup

nathanaelhuffman:
This moves the logic for our naive axi-lite fabric block down into a VHDL 2k8-compatible block and then adds a wrapper with the extant interface.  Doing this required adjusting the package names a bit so we could split the 2k19 and 2k9 code so you see the cascading changes in existing IP that pulled in 2k19-compatible packages.

### DIFF
--- a/hdl/ip/vhd/axi_blocks/axilite_if_2k19_pkg.vhd
+++ b/hdl/ip/vhd/axi_blocks/axilite_if_2k19_pkg.vhd
@@ -10,7 +10,7 @@ use ieee.std_logic_1164.all;
 use ieee.numeric_std.all;
 use ieee.numeric_std_unsigned.all;
 
-package axilite_if_pkg is
+package axilite_if_2019_pkg is
    generic (
       addr_width : integer
    );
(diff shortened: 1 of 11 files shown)


hdl/ip/vhd/axi_blocks/axilite_if_2k19_pkg.vhd
```diff
@@ -10,7 +10,7 @@ use ieee.std_logic_1164.all;
 use ieee.numeric_std.all;
 use ieee.numeric_std_unsigned.all;
 
-package axilite_if_pkg is
+package axilite_if_2019_pkg is
```
Aaron-Hartwig, line 13: Something I really want to nip in the bud is cases where our file names do not match the entity or package that they contain. I find that to be unnecessarily inconvenient, particularly given we only ever have a single entity or package in a file.

hdl/ip/vhd/axi_blocks/axilite_if_2k8_pkg.vhd
```diff
@@ -0,0 +1,22 @@
...
+--
+-- Copyright 2025 Oxide Computer Company
+
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+use ieee.numeric_std_unsigned.all;
+
+package axilite_if_2008_pkg is
```
Aaron-Hartwig, line 13: See comment about file vs package/entity naming above.

Aaron-Hartwig (approved):
LGTM aside from a nit about file vs package/entity naming.

Thanks for ironing this out!

hdl/ip/vhd/axi_blocks/axilite_if_2k19_pkg.vhd
nathanaelhuffman: Yeah this was just sloppy work on my part, I had a mix and went through and fixed the conventions without actually fixing the package names in code. I agree with this generally.

PR #290, opened 2025-02-14
Cosmo Hotplug FPGA 1st pass implementation

### LINKED ISSUE #166: Basic Hotplug FPGA Support
This covers adding support for the hotplug FPGA which needs to:

* Speak some combination of PCA9506/9535 as an i2c target from SP5 (likely multiple PCA9506s)
* Provide a SPI interface to the SP for its own interrogation and out of band shenanigans (more in a moment)
* Implement all of the basic discrete logic that we require to power hotplug and generate alerts back to MPIO

Longer term we will want to find different ways to use this for testing and related (simulating hotplug without actually doing hotplug).

nathanaelhuffman:
A fairly decent sized PR, this brings in most of the MVP functionality for the cosmo hotplug FPGA.

There were a couple of tweaks to some of the IP to support making this easier to generate/loop across, but since none of that stuff is used elsewhere really yet I think it's ok to combined here with this PR.

Additionally there were a few things that GHDL complained about so those got cleaned up here as well.

As an overview, this drops in 5 i2c mux endpoints on one i2c bus to the SP, 2 i2c pca I/O expanders to a 2nd i2c bus going to the SP5, and for-generates 10 copies of CEM logic, wiring between the i/o expanders and the CEM slots as required.

This closes #166 (at least with an implementation that we need to test once we have hw!)


nathanaelhuffman:
This current includes a provisional bump of the oss-cad-suite so that CI will run.  Decisions to be made over in https://github.com/oxidecomputer/quartz/pull/291 since bumping the tool version globally isn't a slam dunk

nathanaelhuffman:
> This current includes a provisional bump of the oss-cad-suite so that CI will run, but only for the VHDL-side of the world. Decisions to be made over in #291 since bumping the tool version globally isn't a slam dunk

#291 landed  so this was rebased on that and now has no tool-chain impacts vs what is already in main

Aaron-Hartwig (approved):

PR #293, opened 2025-02-19
make the I2C Controller VC performance configurable

Aaron-Hartwig:
This commit makes the `i2c_controller_vc` performance configurable. It defaults to the  settings as originally designed. The ability to make this operate at a particular speed is useful for me in the SPD mux context where I want to simulate a controller at a certain speed.


nathanaelhuffman (approved):
lgtm

PR #294, opened 2025-02-20
Add an SPD proxy mux

Aaron-Hartwig:
This commit adds the `spd_proxy_top` block whose intention is to house an I2C controller which will attempt to utilize the DIMM bus while a CPU is not using it. When the CPU begins using the bus the internal controller will wrap up what it is doing as quickly as possible and we will attempt to gracefully hand the bus over the CPU in a way that doesn't cause problems.

### DIFF
--- a/hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
+++ b/hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
@@ -293,21 +293,11 @@ begin
                 v.sda_oe    := '0';
 
                 if tx_start then
-                    v.state         := WAIT_TBUF;
-                    v.counter       := STO_TO_STA_BUF_TICKS;
-                    v.count_load    := '1';
-                end if;
-
-            -- Wait out tbuf to ensure STOP/START spacing
-            when WAIT_TBUF =>
-                if sm_count_done then
-                    -- tbuf is always greater than or equal to the START setup requirement, skip to
-                    -- hold
+                    -- Coming back to IDLE after a transaction means we've waited out tbuf, and tbuf
+                    -- is always greater than or equal to the START setup requirement, skip to hold
                     v.state         := START_HOLD;
                     v.counter       := START_SETUP_HOLD_TICKS;
                     v.count_load    := '1';
-                else
-                    v.count_decr    := '1';
                 end if;
 
             when WAIT_REPEAT_START =>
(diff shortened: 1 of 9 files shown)


hdl/ip/vhd/i2c/controller/link_layer/i2c_ctrl_link_layer.vhd
```diff
@@ -433,7 +425,18 @@ begin
...
+                    v.count_load    := '1';
+                    v.sda_oe        := '0';
+                else
+                    v.count_decr    := '1';
+                end if;
+
+            -- Wait out tbuf to ensure STOP/START spacing
+            when WAIT_TBUF =>
+                if sm_count_done then
+                    v   := SM_REG_RESET;
```
Aaron-Hartwig, line 439: For context as to why I moved WAIT_TBUF to be after the STOP states: since we indicate controller business by "if I'm not in IDLE", this keeps the controller from advertising it is ready for another transaction until it can actually execute that transaction. Behavior today is that it will accept the new transaction but then wait for TBUF anyway.

hdl/ip/vhd/vunit_components/i2c_controller/i2c_ctrlr_vc.vhd
```diff
@@ -120,7 +120,7 @@ begin
                 sda <= 'Z';
                 wait for 60 ns; -- need to be longer than 50ns glitch timing
                 sda <= '0';
-                wait for thd_sta * 4;
+                wait for thd_sta;
```
Aaron-Hartwig, line 123: I believe the `* 4` was unintentional so I removed it. This helps me get more realistic bus timing in my simulation.

hdl/projects/cosmo_seq/spd_proxy/spd_proxy_top.vhd
```diff
@@ -0,0 +1,203 @@
...
+        -- FPGA <-> DIMMs
+        dimm_scl_if : view tristate_if;
+        dimm_sda_if : view tristate_if;
+
+        -- FPGA I2C Interface
+        i2c_command         : in cmd_t;
+        i2c_command_valid   : in std_logic;
+        i2c_ctrlr_idle      : out std_logic;
+        i2c_tx_st_if        : view stream8_pkg.st_sink_if;
+        i2c_rx_st_if        : view stream8_pkg.st_source_if;
```
Aaron-Hartwig, line 38: I expect this portion of the interface to change significantly as we do the broader system implementation for this functionality. This was just quick and dirty to get moving.

nathanaelhuffman (approved):
all seems reasonable

PR #296, opened 2025-02-21
adjust top for sapphire rework

Aaron-Hartwig:
I think we want to disconnect the DIMM0 GHIJKL lines since the rework we have could cause two drivers to fight one another (at least on the HPM_TO_SCM bus).


Aaron-Hartwig:
image

nathanaelhuffman (approved):

PR #297, opened 2025-02-22
initial release tooling for buck2 builds

nathanaelhuffman:
Some light docs and this kinda seems to work so fire in the hole.


Aaron-Hartwig (approved):
:shipit:

PR #299, opened 2025-02-28
Bail in synthesis if inferred latches are detected

nathanaelhuffman:


hdl/ip/vhd/i2c/controller/txn_layer/i2c_ctrl_txn_layer.vhd
```diff
@@ -130,6 +130,7 @@ begin
     begin
         v           := sm_reg;
         is_read     := '1' when sm_reg.cmd.op = READ or sm_reg.in_random_read else '0';
+        txd        := (others => '0');
```
Aaron-Hartwig, line 133: Do you want this change in this commit? Its fine if you do, I just wanted to make sure it wasn't an accident.

nathanaelhuffman: I think I'm going to go with yes, as I don't have this merged anywhere else and without it my fancy new bail on inferred latches does exactly that for this CI run since this does cause an inferred latch w/o this fix.

Aaron-Hartwig: Okay. What target is build this? I wasn't aware of it being part of any synthesized code right now.

nathanaelhuffman: Ok maybe I’m just dumb and I looked at the code, recognized the latch and assumed I found this while doing the ci. How did I actually find this originally? Probably debugging in your grapefruit branch :) so maybe this isn’t required here to pass ci. I’ll get this sorted in the am

PR #310, opened 2025-03-17
Refactor axi lite transactions blocks into a common entity

nathanaelhuffman:
This allows us to share the logic that was currently duplicated in every axi target block.  We were startign to have too much copied code in every axi target block and any future fixing or maintenance was going to get out of hand quickly.

Additionally I slightly re-structured the BUCK files so that rather than importing a package, we import more stuff when bringing in axi-lite stuff so the downstream blocks have fewer common axi-lite parts to manage.


Aaron-Hartwig (approved):
Thank you!

PR #311, opened 2025-03-18
make SDA aribtration logic its own module

Aaron-Hartwig:
This could be parameterized further to be a more general way to arbitrate a shared bus where the FPGA is a proxy, but for now this is what it is.

### DIFF
--- a/hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
+++ b/hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
@@ -1,0 +1,89 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2025 Oxide Computer Company
+--
+-- This block monitors two I2C SDA signals, `a` and `b`, with the intention thatt hey are
+-- essentially connected but proxied by the FPGA. It assumes the inactive state for the bus is high,
+-- and then monitors for `a` or `b` to pull the line low. It will then grant whichever side pulled
+-- the line low the bus until that side releases it, at which point after HYSTERESIS_CYCLES both
+-- sides will be resambled again.
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+entity sda_arbiter is
+    generic (
+        HYSTERESIS_CYCLES : integer
+    );
+    port (
+        clk     : in std_logic;
+        reset   : in std_logic;
+
+        a       : in std_logic;
+        b       : in std_logic;
+        enabled : in std_logic;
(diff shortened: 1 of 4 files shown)


hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
```diff
@@ -0,0 +1,89 @@
...
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2025 Oxide Computer Company
+--
+-- This block monitors two I2C SDA signals, `a` and `b`, with the intention thatt hey are
+-- essentially connected but proxied by the FPGA. It assumes the inactive state for the bus is high,
+-- and then monitors for `a` or `b` to pull the line low. It will then grant whichever side pulled
+-- the line low the bus until that side releases it, at which point after HYSTERESIS_CYCLES both
+-- sides will be resambled again.
```
nathanaelhuffman, line 11:
```suggestion
-- sides will be resampled again.
```

hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
```diff
@@ -0,0 +1,89 @@
...
+end entity;
+
+architecture rtl of sda_arbiter is
+    type sda_control_t is (
+        NONE,
+        GRANT_A,
+        GRANT_B
+    );
+    signal sda_state : sda_control_t;
+    signal hysteresis_cntr : natural;
```
nathanaelhuffman, line 40: In general when using naturals that are not constant, I'd like to see a range applied. This helps the compiler figure out how many bits are necessary.  Some tools will also require a default value for some reason, quartus did this with integers and would use the default instead of your reset value.

hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
```diff
@@ -0,0 +1,89 @@
...
+begin
+
+    a_grant <= '1' when sda_state = GRANT_A else '0';
+    b_grant <= '1' when sda_state = GRANT_B else '0';
+
+    arbitration : process(clk, reset)
+        variable sda_next : sda_control_t;
+    begin
+        if reset then
+            sda_state       <= NONE;
```
nathanaelhuffman, line 50: hysteresis_cntr has no reset

nathanaelhuffman (approved):
Check a few notes there, especially the missing reset as I think that's important to fix. Otherwise seems fine

hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
nathanaelhuffman: Since you have a generic, you can probably just inherit its range.

PR #334, opened 2025-04-23
A bunch of work bringing up cosmo.

nathanaelhuffman:
This is kind of a mess but I'd like to get it all merged in soonish so we can start doing more targetted PRs.
I do want to chat about the failing testcase given my changes to the arbiter.


hdl/projects/cosmo_seq/spd_proxy/spd_proxy_top.vhd
```diff
@@ -242,7 +244,7 @@ begin
 
     sda_arbiter_inst: entity work.sda_arbiter
         generic map(
-            HYSTERESIS_CYCLES => DIMM_I2C_TSP_CYCLES + 7 -- 7 is a bit of a swag given Ruby testing
+            HYSTERESIS_CYCLES => DIMM_I2C_TSP_CYCLES + 500 -- 7 is a bit of a swag given Ruby testing
```
Aaron-Hartwig, line 247: We should probably leave a bit of a theory of operations comment here as to where this comes from. Or maybe just make a new constant that has that comment by it. I know iterating on experiments here is not the highest priority.

hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
```diff
@@ -37,53 +37,62 @@ architecture rtl of sda_arbiter is
...
-                            if a = '0' and b = '1' then
-                                sda_next    := GRANT_A;
-                            elsif a = '1' and b = '0' then
-                                sda_next    := GRANT_B;
+                            -- We were came from GRANT_A so a was high.
+                            if last_grant = GRANT_A and a = '0' then
+                                sda_state    <= GRANT_A;
+                            -- We were came from GRANT_B so B was high
+                            elsif last_grant = GRANT_B and b = '0' then
+                                sda_state    <= GRANT_B;
```
Aaron-Hartwig, line 63: Could you elaborate a bit on why this came to be? I just want to make sure I understand the reasoning. I'm happy to get sim coverage here in parallel to your other efforts.

hdl/projects/cosmo_seq/sequencer/a1_a0_seq.vhd
```diff
@@ -86,7 +94,7 @@ architecture rtl of a1_a0_seq is
         ddr_bulk_en: std_logic;
         group_a_en: std_logic;
         group_b_en: std_logic;
-        group_c_en: std_logic;
+        group_c_en: std_logic_vector(3 downto 0);
```
Aaron-Hartwig, line 97: Having this be an slv without named constants for the indices (or a record) feels prone to error. Probably not a big deal at this point if things are working, but as it is currently written you don't really know what is going by reading the state machine, you'd have to go look at what each bit is driving ~200 lines down.

hdl/projects/cosmo_hp/hp_subsystem/sims/cem_hp_sim_th.vhd
```diff
@@ -0,0 +1,193 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2024 Oxide Computer Company
```
Aaron-Hartwig, line 5:
```suggestion
-- Copyright 2025 Oxide Computer Company
```

hdl/projects/cosmo_hp/hp_subsystem/sims/cem_hp_sim_tb.vhd
```diff
@@ -0,0 +1,58 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2024 Oxide Computer Company
```
Aaron-Hartwig, line 5:
```suggestion
-- Copyright 2025 Oxide Computer Company
```

hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
nathanaelhuffman: So when looking at the original arbiter, it was a bit confusing to me because it would run hysteresis any time there was a selection at all and it seemed to make more sense to me to only have hysteresis when you were changing grants?  I'm not married to this but this made more sense to me as I went through trying to understand what was going on here.

It is possible that I don't have the full picture or am missing something also :D

hdl/projects/cosmo_seq/sequencer/a1_a0_seq.vhd
nathanaelhuffman: This was left over from debugging. I've made all the c rails sequence concurrently now and removed this hack so there's only one group_c_en bit again like there was originally. thanks!

hdl/projects/cosmo_hp/hp_subsystem/sims/cem_hp_sim_th.vhd
nathanaelhuffman: fixed

hdl/projects/cosmo_seq/spd_proxy/spd_proxy_top.vhd
nathanaelhuffman: I pushed a fix reverting this change

hdl/projects/cosmo_seq/spd_proxy/sda_arbiter/sda_arbiter.vhd
Aaron-Hartwig: So the reason I had it running hysteresis every time the grant changed is because I also consider going from one grant to no grants as a "change" (both sides are now high Z). I'm open to discussion on if we should have memory here to fast-track re-granting the bus, but my gut reaction is that we should not do that given the speed SDA _should_ be changing.

Aaron-Hartwig: For future us: we're just going to adjust the sims to account for the new behavior.

nathanaelhuffman:
This was a mis-merge and I'm bad at git.

PR #336, opened 2025-04-23
minibar: adjustments to refclk buffer defaults, typo fixes

Aaron-Hartwig:
Changes to the PCIE refclk buffer control defaults:
- bw_sel: value of `2'h1` which is "PLL with High Bandwidth" (see https://github.com/oxidecomputer/quartz/blob/main/hdl/projects/minibar/MinibarPcie.bsv#L108-L130 for how this register decodes to the tristate pin)
- pd: changed to be asserted by default, this was a miss from earlier
- oe1: changed to enabled by default

The plan is that after programming the FPGA, the SP will come and clear the `pd` register.

The other changes are just register description typos I missed earlier.


Aaron-Hartwig:
This change has been validated on my Minibar with a 1G NIC in the PCIe slot and functional to the Gimlet.

nathanaelhuffman (approved):

PR #337, opened 2025-04-28
Fix nic hotplug logic and MAPO if CPU asserts SMERR_L

nathanaelhuffman:


hdl/projects/cosmo_seq/cosmo_timing.xdc
```diff
@@ -96,11 +96,13 @@ create_clock -name sp5_espi_clk_ext -period 15.000;
...
 # Data 
 # max= 7.5ns (1/2 period) + 3ns (maxreal tco)
 # min= 7.5ns (1/2 period) + 1ns (min real tco)
-set_input_delay -clock sp5_espi_clk_ext -max 10.5 [get_ports espi0_sp5_to_fpga1_dat[*]]
-set_input_delay -clock sp5_espi_clk_ext -min 8.5 [get_ports espi0_sp5_to_fpga1_dat[*]]
\ No newline at end of file
+#set_input_delay -clock sp5_espi_clk_ext -max 10.5 [get_ports espi0_sp5_to_fpga1_dat[*]]
+#set_input_delay -clock sp5_espi_clk_ext -min 8.5 [get_ports espi0_sp5_to_fpga1_dat[*]]
+set_max_delay -to [get_ports espi0_sp5_to_fpga1_dat[*]] 6
+set_min_delay -to [get_ports espi0_sp5_to_fpga1_dat[*]] 0
```
Aaron-Hartwig, line 108: Maybe throw a comment up top here about the goals of the timing constraints here?

Aaron-Hartwig (approved):

hdl/projects/cosmo_seq/cosmo_timing.xdc
Aaron-Hartwig: I mean this comment to apply to the whole eSPI block here. I think we should explain what the goal was with the commented out stuff (you give the math already) and then the why of the two lines you've added here.

nathanaelhuffman: I don't disagree, but I'm going to rewrite this and properly time it so this is mostly patches on patches until I do the right design here which I have out on paper but not coded yet.

nathanaelhuffman: I'll just remove the commented out crap and make a note regardless

Aaron-Hartwig: That works for me! The current state gives WIP vibes, which it is, but I know that because we talked about it already haha.

PR #339, opened 2025-05-01
Add byte-enables to DPR

nathanaelhuffman:
I have a use-case that wants byte-enables on a dual-port ram so I'm adding them and an associated testcase here.
@Aaron-Hartwig: Ready to review

### DIFF
--- a/hdl/ip/vhd/memories/mixed_width_simple_dpr.vhd
+++ b/hdl/ip/vhd/memories/mixed_width_simple_dpr.vhd
@@ -1,0 +1,132 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2025 Oxide Computer Company
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+use ieee.numeric_std_unsigned.all;
+
+use work.calc_pkg.all;
+
+-- Simple dual port ram supporting dual clocks with symmetric read and
+-- write sizes. Addresses are in terms of `WIDTH` as a word
+-- `NUM_WORDS` is the depth of the FIFO again in terms of entries
+-- of `WIDTH` of the port
+-- Note: a read from an address that is undergoing a write will
+-- result in undefined behavior and needs to be avoided
+-- Note2: This is an attempt at making a ram that can be inferred
+-- by vivado wihout using shared variables. All of Xilinx's inferrence
+-- examples use shared variables who's functionality was deprecated
+-- technically by VHDL-2002, and a protected variable would be a more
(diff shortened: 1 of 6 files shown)


hdl/ip/vhd/memories/mixed_width_simple_dpr.vhd
```diff
@@ -0,0 +1,113 @@
+library ieee;
```
Aaron-Hartwig, line 1: nit: paste in the usual header

hdl/ip/vhd/memories/mixed_width_simple_dpr.vhd
```diff
@@ -0,0 +1,113 @@
...
+        waddr : in    std_logic_vector(log2ceil(write_num_words) - 1 downto 0);
+        -- Write data, sync'd to wclk domain
+        wdata : in    std_logic_vector(write_width - 1 downto 0);
+        -- Write enable, sync'd to wclk domain
+        wren : in    std_logic; -- wclk domain
+        -- Read-side interface clock
+        rclk : in    std_logic;
+        -- Read address, sync'd to rclk domain
+        -- Worked example: write_w = 8, write_depth = 16, read_w = 16, read_depth = 8
+        -- read addrss size is going to be
```
Aaron-Hartwig, line 28: Seems like an incomplete thought. Perhaps this sort of theory of operations should go in a block comment at the top?

hdl/ip/vhd/memories/mixed_width_simple_dpr.vhd
```diff
@@ -0,0 +1,113 @@
...
+                    -- Set up the address by concatenating the ratio bits to the end since we're accessing
+                    -- by min-width sizes.
+                    addr := to_integer(waddr);
+
+                    if wren = '1' then
+                        ram(addr) <= wdata;
+                    end if;
+                end loop;
+            end if;
+        end process;
```
Aaron-Hartwig, line 87:
```suggestion
        wsmaller_write_proc:process(wclk)
        begin
            if rising_edge(wclk) then
                for i in 0 to RATIO - 1 loop
                    -- Set up the address by concatenating the ratio bits to the end since we're accessing
                    -- by min-width sizes.
                    if wren = '1' then
                        ram(to_integer(waddr)) <= wdata;
                    end if;
                end loop;
            end if;
        end process;
```
nit: Should we just nix the `addr` variable in this process? I like it for when we have a more involved address generation function that would be messy of have on the LHS (like up on L59 or down on L95), but the write address is simple for this case.

hdl/ip/vhd/memories/mixed_width_simple_dpr.vhd
Aaron-Hartwig: It may also be worth mentioning what the intention is for a "simple" mixed-width DPR. i.e. the bigger side is going to just be as parallel as needed to satisfy the ratio, there's no latency-based shenanigans.

PR #351, opened 2025-05-14
Add 8b10b encoders/decoders 

nathanaelhuffman:
based on chuck benz verilog example.

Added some helper functions which are expected to be useful for ignition simulations where this will be properly exercised, but this unit felt stand-alone enough to be it's own foundational PR.

Rather than build out a bunch of simulations for these blocks  (which are effectively just math functions) I'd  rather focus on integrating this stuff into a more useful ignition simulation and use those test  vehicles to be the ongoing tests of these modules.

### DIFF
--- a/hdl/ip/vhd/8b10b/decode_8b10b.vhd
+++ b/hdl/ip/vhd/8b10b/decode_8b10b.vhd
@@ -56,7 +56,7 @@ architecture rtl of decode_8b10b is
 
     signal disp6b : std_logic;
 
-    -- The 5B/6B decoding special cases where ABCDE !       : std_logic; abcde
+    -- The 5B/6B decoding special cases where ABCDE != abcde
 
     signal p22bceeqi   : std_logic;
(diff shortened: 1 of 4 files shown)


hdl/ip/vhd/8b10b/decode_8b10b.vhd
```diff
@@ -0,0 +1,219 @@
...
+    signal p40 : std_logic;
+    signal p04 : std_logic;
+
+    signal disp6a  : std_logic;
+    signal disp6a2 : std_logic;
+    signal disp6a0 : std_logic;
+
+    signal disp6b : std_logic;
+
+    -- The 5B/6B decoding special cases where ABCDE !       : std_logic; abcde
```
Aaron-Hartwig, line 59:
```suggestion
    -- The 5B/6B decoding special cases where ABCDE != abcde
```
this got mangled by find and replace it looks like

Aaron-Hartwig (approved):

PR #356, opened 2025-05-15
Fix helper function encoded bit order

nathanaelhuffman:
Due to how the 8b10b algo works, most tables show the input (8b) in octal or hex with MSB 1st (HGF EDCBA) but show the encoded 10b version in LSB1st (abcdei fghj). Flipping the bit-order in the constants makes it more difficult to compare to tables for correctness, so I have left things in the encoded order, but changed the names and added comments to reflect that and then reversed the bit order before it's consumed.

Problems like this are extremely annoying to notice/debug so it's best to give ourselves the easiest time possible.

This also adds a helper function that will figure out control or non-control character encoding.


Aaron-Hartwig (approved):
I agree with the decision to make the things we reference be presented in the form they generally tend to be referred to in, and then do the reversal at the end.

PR #357, opened 2025-05-17
Add/move some crc functions into a crc_helpers package

nathanaelhuffman:
These are useful for testbenching and it makes more sense to put them here with the crcs than it does having them spread out in different places.
This consolidates the crc8_atm from the espi block, fixing the espi sims, and adds a function of the same shape for the autostar which we'll use in ignition sims.


Aaron-Hartwig (approved):
LGTM. Looks like you need to add `crc_sim_pkg` to your `hdl/ip/vhd/crc/BUCK` to fix sims.

PR #358, opened 2025-05-20
Make SP-console the default UART

nathanaelhuffman:
Previously, we defaulted to shoving the SP5 uart data out the debug header. That presented challenges with the mfg stations where a uart dongle isn't necessarily connected leading to a wedge due to hardware-handshake.  This makes the default the  SP which will keep the buffer clear (like it does on gimlet) and we have a new memory mapped peripheral to flip the mux back to the debug headers if we need it.

This lays the groundwork for additional debug control of various things which we may want eventually also.

I booted this on my cosmo, and once booted I could use the `humility console-proxy attach` command to get proper uart coms. This verified that we don't wedge during boot and the the sp-console-proxy is working as expected.

### DIFF
--- a/hdl/projects/cosmo_seq/sp5_uart_subsystem/sp5_uart_subsystem.vhd
+++ b/hdl/projects/cosmo_seq/sp5_uart_subsystem/sp5_uart_subsystem.vhd
@@ -9,12 +9,14 @@ use ieee.std_logic_1164.all;
 use ieee.numeric_std.all;
 
 use work.axi_st8_pkg.all;
+use work.debug_regs_pkg.all;
 
 entity sp5_uart_subsystem is
     port(
         clk : in std_logic;
         reset : in std_logic;
 
+        dbg_uart_control : in uart_control_type;
         -- sp UART pins, ok to be un-syncd
         ipcc_from_sp : in std_logic;
         ipcc_to_sp : out std_logic;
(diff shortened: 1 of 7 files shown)


hdl/projects/cosmo_seq/sp5_uart_subsystem/sp5_uart_subsystem.vhd
```diff
@@ -66,11 +67,21 @@ begin
...
+                -- outputs to host, inputs from debug header
+                host_from_fpga <= dbg_pins_uart_in;
+                host_from_fpga_rts_l <= dbg_pins_uart_out_rts_l;
+            else
+                -- Send output of the host uart to the host bits.
+                host_from_fpga <= fgpa_sp_to_host_int;
+                host_from_fpga_rts_l <= fpga_sp_to_host_int_rts_l;
+                dbg_pins_uart_out <= '1';
+                dbg_pins_uart_in_rts_l <= '1';
+            end if;
```
Aaron-Hartwig, line 84: Do we want to give them a known value when `sp5_to_header` is `'0'`?

Aaron-Hartwig (approved):
One question, otherwise :shipit:

hdl/projects/cosmo_seq/sp5_uart_subsystem/sp5_uart_subsystem.vhd
nathanaelhuffman: At least for now yes since for the foreseeable future we have a UART here and '1' is IDLE and FULL  respectively.

I'd eventually like to have some fancier stuff here but I'm intending on leaving this header generally dedicated to UART until cosmo rev2 where we'll have a dedicated header for UART things.

nathanaelhuffman: also note that without doing something like this, we'd infer a latch on those signals since this is in a clocked process.

PR #394, opened 2025-07-17
implement the oximux16 in front and rear fpgas

nathanaelhuffman:
This was tested in hardware with a special branch from matt and things look good.

I also included some additional SP5 debugging stuff for pwrgd_out that was pending.

Additionally I had a hot timing path in espi that I was able to add a register for to gain back some margin w/o a functional effect.


nathanaelhuffman:
I decided to merge this so I could push to hubris.

PR #403, opened 2025-07-29
Re-do fmc timing constraints

nathanaelhuffman:
new answers here :)


Aaron-Hartwig (approved):
I believe this looks correct. My one ask would be that rather than having "magic numbers" in the delay calculations that you instead assign all numbers a name that explains what it is (and ideally where it is defined or how it was derived). You list this in comments, but then will use a number in the formula. I think constrain calculations are easier to reason about when you can read them almost like a sentence.

PR #407, opened 2025-08-01
cosmo-seq: fix missing wire for sequencer IRQ

Aaron-Hartwig:
We got it into the register interface and then from the sequencer to the pin, but didn't tie the register to the pin in the sequencer itself.


nathanaelhuffman (approved):
Oops, TY!

PR #408, opened 2025-08-02
Allow remapping APOB flash accesses to the bonus flash area

citrus-it:
Undo the mapping of flash address 0x1_000_000 to the bonus region at
0x4_000_000 and replace instead with a configurable range set up via
two new registers. This allows the SP to arrange things so that reads
to the APOB region of the SPI image are redirected to the bonus flash
where it has previously stashed an APOB from a past successful boot.

---------

Co-authored-by: Aaron Hartwig <aaron@oxidecomputer.com>


Aaron-Hartwig (approved):
I am good with merging this but will leave the final approval to Nathanael just to make sure he's good with things.

nathanaelhuffman (approved):

PR #411, opened 2025-08-07
pmbus sequencer alerts

nathanaelhuffman:
We bring a bunch of alert signals from on-board sm/pmbus regulators into the interrupt block on cosmo so hubris can choose to get real interrupts (or poll the live status) on these.

We also turn them to be active high in the IFR register since that matches other interrupt sources.


Aaron-Hartwig (approved):
Exciting! And thanks for explaining which rails correspond to the various `pwr_cont$` in the RDL.

PR #413, opened 2025-09-17
cosmo_seq: add ApobFlashOffset register

Aaron-Hartwig:
Implemented per https://rfd.shared.oxide.computer/rfd/593#_apob_flash_offsets and discussions with Matt.

### DIFF
--- a/hdl/ip/vhd/spi_nor_controller/espi_txn/espi_flash_txn_mgr.vhd
+++ b/hdl/ip/vhd/spi_nor_controller/espi_txn/espi_flash_txn_mgr.vhd
@@ -21,6 +21,7 @@ entity espi_flash_txn_mgr is
         cur_flash_addr_offset: in signed(31 downto 0);
         cur_apob_flash_addr: in std_logic_vector(31 downto 0);
         cur_apob_flash_len: in std_logic_vector(31 downto 0);
+        cur_apob_flash_offset: in std_logic_vector(31 downto 0);
         -- espi cmd fifo interface
         espi_cmd_fifo_rdata: in std_logic_vector(31 downto 0);
         espi_cmd_fifo_rdack: out std_logic;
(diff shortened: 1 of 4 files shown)


hdl/ip/vhd/spi_nor_controller/espi_txn/espi_flash_txn_mgr.vhd
```diff
@@ -117,7 +118,7 @@ begin
                    unsigned(r.cur_flash_addr) <  unsigned(r.apob_end_addr)then
                     -- Remap to raw address space starting at 0x4000000; the bonus flash region
                     v.cur_flash_addr := std_logic_vector(
-                        to_unsigned(16#4000000#, 32) +
+                        unsigned(cur_apob_flash_offset) +
```
citrus-it, line 121: Feel free to ignore - I wasn't in the the conversation you had with @mkeeter - I had expected this would be an offset into the bonus flash region rather than an absolute offset into flash. Either works but perhaps update the comment at 119.

hdl/ip/vhd/spi_nor_controller/spi_nor_regs.rdl
```diff
@@ -122,25 +122,33 @@ addrmap spi_nor_regs {
...
+        desc = "Allows the SP to configure the length of the flash range which is remapped to APOB flash region";
         field {
             desc = "Length of the APOB flash region";
-        } offset[31:0] =  0;
+        } offset[31:0] = 0;
     } ApobFlashLen;
+
+    reg {
+        name = "APOB Flash offset address";
+        desc = "Allows the SP to configure the offset for which slot to access in the APOB flash region";
```
citrus-it, line 149: How about something like:

```suggestion
        desc = "Allows the SP to configure the selected APOB region via its absolute offset info flash";
```

PR #426, opened 2025-10-21
put ignition rev2+ builds into play

nathanaelhuffman:
This puts the changes in place for rev2+ cosmo ignition images. This also creates a new fpga build for rev1 which is essentially orphaned due to different pin config.

In building this, we had a timing failure in oximux 16 on the cosmo-hp build so I adjusted the 1hot detection logic to be more logic-efficient.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,7 @@ jobs:
       cosmo_hp: ${{ steps.filter.outputs.cosmo_hp }}
       cosmo_seq: ${{ steps.filter.outputs.cosmo_seq }}
       cosmo_ignition: ${{ steps.filter.outputs.cosmo_ignition }}
+      cosmo_ignition_a: ${{ steps.filter.outputs.cosmo_ignition_a }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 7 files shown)


hdl/projects/cosmo_ignition/cosmo_ignition_r1_top.vhd
```diff
@@ -0,0 +1,84 @@
...
+    );
+end entity;
+
+architecture rtl of cosmo_ignition_r1_top is
+
+begin
+
+    cosmo_ignition_top_inst: entity work.cosmo_ignition_top
+     generic map(
+        IS_REV1 => true
```
Aaron-Hartwig: Ah, I see we _do_ need to keep that rev1 build around. I was thinking we had version resistors here.

Aaron-Hartwig (changes requested):
Rather than comment throughout I'll just leave a top level suggestion. We should make the naming here explicit about which version of boards will be supported (we can debate rev numbers vs HCV letters). But I don't think having an _r1 build and a build without any suffixes makes it immediately clear what is supported. On other boards we've done things like add letters when the same image supports multiple versions, e.g., `_abc.bit`.

.github/workflows/build.yml
```diff
@@ -113,6 +113,28 @@ jobs:
             path: "./buck-out/v2/gen/root/**/*"
         - name: Cleanup
           run: bash .github/cleanup.sh
+  cosmo_ignition_r1_only:
```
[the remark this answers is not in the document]
nathanaelhuffman: i'm ok with this and will fix tomorrow

hdl/projects/cosmo_ignition/cosmo_ignition_r1_top.vhd
nathanaelhuffman: Yes given the changes, rev1 becomes a special case, and all future cosmos need a different image.

PR #432, opened 2025-10-31
espi post code ring buffer

### LINKED ISSUE #419: Cosmo Sequencer: Want a buffer of post codes
We currently expose the last post code observed today, but it would be helpful to have a buffer of post codes the system has observed throughout the boot process. The initial thought here was to have a ring-buffer of post codes, but I think we are better served by just implementing this as a standard, deep buffer. This is because post codes either change very infrequently or a lot all at once, so deciding on how many previous codes we care about could be unnecessarily limiting since we have plenty of available BRAM (only 11% utilization) in this design. We'd also need to expand the internal AXI bus address space to accommodate addressing this buffer. 

@luqmana thought that 2048 would likely be workable (or 4096 to be on the safe side) given he's observed on the order of ~1500 codes being posted during various boots:
```
$ <turin-ruby-1.0.0.7-postcodes.txt | wc -l
1453
$ <turin-ruby-1.0.0.7-postcodes.txt | uniq | wc -l
1428
```

The idea would be to just have a deep enough buffer that we could record the entire history of events and not worry about "ring-buffer" things like rolling over and generation counts. This is not a big deal from the FPGA's perspective, but we need to think through the rest of the API and plumbing of how this information gets to the host in a useful format. A few questions I have are:

1. What are our options for bursting reads via FMC<->AXI? 4096 entries would be a lot to read one at a time.
2. How would the SP store the intermediate buffer of this data? Can we just store it in RAM? 4096 * 32-bits = 16KiB
3. How does this get passed along to the host?

nathanaelhuffman:
This ended up being a bit of an adventure. The implementation of the buffer was actually very straight forward, but we had not to this point, supported mixed size peripherals so we had to deal with that as well.

The solution represented here is functional but not optimal: This makes the "responders" interface from the fabric all equivalently sized to 32 bit and the implementer needs to "trim" the axil interface down at point of use. This allows the fabric to use loops over the same-size datatypes and we can make overloaded procedures to trim them down to the appropriate size for a given peripheral. This does foist some work on the implementer so it's not quite as clean and plug-and-play as we'd like- such is the cost of complexity.

The [axilite_if_2k19_helper_pkg.vhd](https://github.com/oxidecomputer/quartz/compare/ndh/cosmo-postcode-buffer?expand=1#diff-bab8a559edf732b63a346c8e58be537f421eb396bced9a6842f501f3f918c42d) contains examples of the overloading of trimming providers. I couldn't quite get generic procedures + views working otherwise this could have been a single generic procedure, as it is, you have to copy-paste it for each target size you want to support :/

Then we implement the post-code stuff: we keep the "last seen" register but add a counter and the buffer.  Both counter and buffer pointers are reset on an espi-reset, so they remain useful even after an A0->A2 transition, only resetting at the beginning of the AMD espi boot sequence when the espi-reset is issued. (The buffer is ram, so its contents aren't zeroed).

With some hubris patches to support the memory array, this is working on my cosmo A:

The counter counts during boot:
```
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008024
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0x491
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008024
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0x4ac
```

last-seen is stable as before (we're booted at this point so no more post-codes):
```
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008020
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0xee110005
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008020
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0xee110005
```

spot-checking buffer entries shows presumably valid payloads:
```
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008100
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0xee1000b2
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008104
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0xee1000a0
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008108
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0xee1000a3
```

And a full boot sequence lands at about 0x51f codes:
```
➜  hubris git:(d5a8d559) ✗ humility hiffy -c FmcDemo.peek32 -a addr=0xc0008024
humility: attached via CMSIS-DAP
FmcDemo.peek32() => 0x51f
```

I took the unit to A2, verified none of this stuff changed and the booted to A0 again and by the time I could look at it, we had some post-codes, but a lower number than 0x51f showing that the counter resets properly.

Fixes #419.

### DIFF
--- a/hdl/ip/vhd/espi/sims/espi_tb.vhd
+++ b/hdl/ip/vhd/espi/sims/espi_tb.vhd
@@ -251,7 +251,11 @@ begin
                 read_bus(net, bus_handle, To_StdLogicVector(POST_CODE_COUNT_OFFSET, bus_handle.p_address_length), data_32);
                 check_equal(data_32, std_logic_vector'(x"00000001"), "Post code count register readback failed");
 
-                -- TODO: issue an espi reset and verify post code count resets
+                -- issue an espi reset and verify post code count resets
+                dbg_espi_reset(net);
+                wait for 1 us;
+                read_bus(net, bus_handle, To_StdLogicVector(POST_CODE_COUNT_OFFSET, bus_handle.p_address_length), data_32);
+                check_equal(data_32, std_logic_vector'(x"00000000"), "Post code count register did not reset after espi reset");
 
 
             elsif run("advanced_post_code_check") then
@@ -292,8 +296,11 @@ begin
                 read_bus(net, bus_handle, To_StdLogicVector(POST_CODE_COUNT_OFFSET, bus_handle.p_address_length), data_32);
                 check_equal(data_32, std_logic_vector'(x"00000002"), "Post code count register readback failed");
 
-                -- TODO: issue an espi reset and verify post code count resets
-
+                -- issue an espi reset and verify post code count resets
+                dbg_espi_reset(net);
+                wait for 1 us;
+                read_bus(net, bus_handle, To_StdLogicVector(POST_CODE_COUNT_OFFSET, bus_handle.p_address_length), data_32);
+                check_equal(data_32, std_logic_vector'(x"00000000"), "Post code count register did not reset after espi reset");
 
             elsif run("msg_w_data_uart") then
                 enable_debug_mode(net);
--- a/hdl/ip/vhd/axi_blocks/axil8_resizer.vhd
+++ b/hdl/ip/vhd/axi_blocks/axil8_resizer.vhd
@@ -1,5 +1,3 @@
-
-
 -- This Source Code Form is subject to the terms of the Mozilla Public
 -- License, v. 2.0. If a copy of the MPL was not distributed with this
 -- file, You can obtain one at https://mozilla.org/MPL/2.0/.
--- a/hdl/ip/vhd/espi/espi_target_top.vhd
+++ b/hdl/ip/vhd/espi/espi_target_top.vhd
@@ -101,6 +101,8 @@ architecture rtl of espi_target_top is
     signal post_code_valid : std_logic;
     signal espi_reset_strobe : std_logic;
     signal espi_reset_strobe_syncd : std_logic;
+    signal espi_link_layer : std_logic;
+    signal dbg_espi_reset : std_logic;
 
 begin
 
@@ -173,7 +175,7 @@ begin
         wait_states => wait_states_fast,
         qspi_mode => encode(qspi_mode_vec_fast),
         alert_needed => alert_needed_fast,
-        espi_reset => espi_reset_strobe
+        espi_reset => espi_link_layer
     );
 
     -- debug_link_layer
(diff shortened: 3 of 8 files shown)


hdl/ip/vhd/espi/sims/espi_tb.vhd
```diff
@@ -232,6 +232,68 @@ begin
...
+                -- Check post code buffer entry 0
+                read_bus(net, bus_handle, To_StdLogicVector(POST_CODE_BUFFER_OFFSET, bus_handle.p_address_length), data_32);
+                check_equal(data_32, std_logic_vector'(x"000001de"), "Post code buffer 0 readback failed");
+                read_bus(net, bus_handle, To_StdLogicVector(POST_CODE_BUFFER_OFFSET + 4, bus_handle.p_address_length), data_32);
+                check_equal(data_32, exp_data_32, "Post code buffer 1 readback failed");
+                -- Check post code count register
+                read_bus(net, bus_handle, To_StdLogicVector(POST_CODE_COUNT_OFFSET, bus_handle.p_address_length), data_32);
+                check_equal(data_32, std_logic_vector'(x"00000002"), "Post code count register readback failed");
+
+                -- TODO: issue an espi reset and verify post code count resets
```
Aaron-Hartwig: Do you want to address this prior to merging?

hdl/projects/cosmo_seq/cosmo_seq_top.rdl
```diff
@@ -14,10 +14,10 @@ addrmap cosmo_seq_top {
     // Instantiate 2 to test nesting
     info_regs info @ 0x0;
     spi_nor_regs spi_nor @ 0x0100;
-    espi_regs espi @ 0x0200;
-    sequencer_regs sequencer @ 0x300;
-    sp_i2c_regs sp_i2c @ 0x0400;
-    pca9506_axi_regs fpga1_hotplug @ 0x0500;
-    dimm_regs dimms @ 0x0600;
-    debug_regs debug_ctrl @ 0x0700;
```
Aaron-Hartwig: Maybe just a future piece of food for thought, but should we make the default space we are allocating bigger than 8-bits wide? Like perhaps defaulting to peripherals with 12 or 16-bit address spaces?We can't see the future, and 256-bytes has been fine for all sorts of things, but as soon as you want a RAM it all blows up. This is not an immediately actionable thought.

hdl/ip/vhd/espi/sims/espi_tb.vhd
nathanaelhuffman: ugh. Not really as it means I need to adjust both the espi control simulation model and it's not really designed to do this :/  I'll think about how I might do something here in a more easy-to-implement fashion for now. I did test this functionality in hardware explicitly.

PR #436, opened 2025-11-13
Fix i2c caching bugs around engine restart

nathanaelhuffman:
- Bug # 1: we latch the current dimm # and restart, thus skipping dimm0 on restart.
- Bug # 2: On an engine restart, we didn't set the page back to 0. This was actually tricky as we *intended* to do that but on a restart the i2c engine could be in a completed state and we'd skip the first page-set transaction accidentally.


Aaron-Hartwig (approved):
Glad to have these cache bugs ironed out!

PR #489, opened 2026-04-14
Bump vunit version

sdonnan:
As is the `tools/requirements.txt` does not work on a fresh Ubuntu 24.04 install due to the deprecation and removal of `pkg_resources`. This fixes the issue and removes unused `json4vhdl` feature that is now handled differently in `vunit`.


nathanaelhuffman (approved):
lgtm

PR #491, opened 2026-04-17
Monitor key postcodes

### LINKED ISSUE #490: Record and expose whether key POST codes have been seen during boot
There are a few POST codes that show up just once per boot on cosmo where it would be useful
to know if they have been seen since the last transition to A0. This gives a coarse indication of
PSP boot progress that could be interpreted and displayed. Perhaps we could provide a register
with bits that latch true as each of these is seen, or some similar mechanism.

- 0xee1000a0 - BL_SUCCESS_C_MAIN - Successfully entered Main
- 0xea00e046 - TpProcMemAfterMemDataInit - memory data initialised
- 0xea00e101 - TpAbl7ResumeInitialization - eMCR starting (not seen if no eMCR)
- 0xea00e340 - TpAblMemoryDdrTrainingStart - seen instead of ^ when no eMCR
- 0xea00e055 - TpProcCpuOptimizedBootStart
- 0xea00e0c9 - TpAbl4Apob
- 0xee1000bb - BL_SUCCESS_BIOS_LOAD_COMPLETE
- 0x1de90001 - PHBLhello

nathanaelhuffman:
Fixes #490.

New register provided, new test cases.  We may have to think about what happens when we begin sharing the espi-core across multiple AMD cpu families which could presumably have different opcodes, but I think we can probably support that with some compile-time generics if/when the time comes.

### DIFF
--- a/hdl/ip/vhd/espi/sys_regs/sp5_post_code_pkg.vhd
+++ b/hdl/ip/vhd/espi/sys_regs/sp5_post_code_pkg.vhd
@@ -1,0 +1,69 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2025 Oxide Computer Company
+
+-- Composite record and views for exposing eSPI spec register
+-- values as a read-only interface in the sys_regs address space.
+
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+use work.espi_spec_regs_pkg.all;
+use work.espi_regs_pkg.all;
+
+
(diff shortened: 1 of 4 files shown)


hdl/ip/vhd/espi/sys_regs/sp5_post_code_pkg.vhd
```diff
@@ -0,0 +1,69 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2025 Oxide Computer Company
```
Aaron-Hartwig:
```suggestion
-- Copyright 2026 Oxide Computer Company
```

Aaron-Hartwig (approved):

PR #495, opened 2026-05-06
Ndh/spd proxy fixes

### LINKED ISSUE #496: i2c proxy bug1: SDA could while playback SCL is high
In `PLAY_STORED_DATA`, the `playback_scl_redge` branch transitions to `ENSURE_PLAYBACK_HOLD` when `v.playback_bits = sample_r.bit_count` and `cpu_scl_in = '1'`:

```vhdl
-- line 335-345
if playback_scl_redge = '1' then
    v.playback_bits := mux_r.playback_bits + 1;
    if v.playback_bits = sample_r.bit_count and cpu_scl_in = '1' then
        v.state := ENSURE_PLAYBACK_HOLD;
        ...
        v.dimm_sda_oe := not cpu_sda_in;   --  <--maybe problem
    end if;
```

`mux_r.dimm_sda_oe` changes in the cycle _after_ the rising playback SCL edge — but `playback_r.scl_out` will still be '1' (SCL is high). Changing SDA while SCL is high is a start/stop condition to any I2C target watching the bus. The same class of bug possibly exists in the corresponding branch at lines 349–357 (the fedge branch with `cpu_scl_in = '0'` is fine; the redge branch is not).

Possible fix:
only transition to `ENSURE_PLAYBACK_HOLD` on `playback_scl_fedge` branches — never on `playback_scl_redge`. If `cpu_scl_in = '1'` on the rising edge, just let the playback counter run to the next falling edge and re-evaluate. The `ENSURE_PLAYBACK_HOLD` hold timer gives the glitch margin; you don't need to also move the transition to the rising edge.

Note that this is by analysis and sim only, I haven't managed to capture this bug actually occurring in hw, but that doesn't mean it can't or doesn't!

### LINKED ISSUE #497: i2c proxy bug 2: State machine could get stuck, then `playback_bits` overflows the `data_bits` index
In `PLAY_STORED_DATA`, the only exit conditions on `playback_scl_fedge` are:

```vhdl
-- line 346-357
elsif playback_scl_fedge = '1' and mux_r.playback_bits < sample_r.bit_count then
    ...  -- output next bit
elsif playback_scl_fedge = '1' and mux_r.playback_bits = sample_r.bit_count and cpu_scl_in = '0' then
    ...  -- hand off
end if;
```

The case `playback_scl_fedge = '1'` and `mux_r.playback_bits = sample_r.bit_count` and `cpu_scl_in = '1'` is entirely unhandled. Since playback runs considerably faster than the CPU SCL, this is the _common_ case: playback finishes all buffered bits while the CPU SCL is still in its high phase.

What happens: the state machine sits in `PLAY_STORED_DATA` as the playback counter keeps running. On the next `playback_scl_redge`, `v.playback_bits` becomes `bit_count + 1`. Now neither fedge condition ever fires again (`playback_bits < bit_count` is false; `playback_bits = bit_count` is false). The machine is stuck, and each subsequent `playback_scl_redge` increments `playback_bits` further. Eventually it indexes `sample_r.data_bits(7)` or higher — out of bounds on a `unsigned(6 downto 0)`.

**Possible Fix**: when playback has caught up (`mux_r.playback_bits >= sample_r.bit_count`), the right trigger for handoff is the CPU's own falling edge (`cpu_scl_fedge`), not a playback falling edge. Add a branch:

```vhdl
elsif cpu_scl_fedge = '1' and mux_r.playback_bits >= sample_r.bit_count then
    v.state := ENSURE_PLAYBACK_HOLD;
    ...
end if;
```

This also eliminates the need to check `cpu_scl_in` on the playback edges for handoff: just wait for the actual CPU SCL falling edge.

Also by analysis and simulation, I haven't caught this in real hardware, but it is a class of bug totally dependent on data patterns and timing

### LINKED ISSUE #498: i2c proxy bug3: invalid start hold time on handoff
`PLAY_STORED_START` has an immediate handoff that bypasses start-condition hold time

```vhdl
-- line 310-312
if sample_r.state = SAMPLE_START and cpu_scl_in = '1' then
    v.state := CPU_HAS_BUS;
    v.playback_done := '1';
```

`CPU_HAS_BUS` immediately clears `dimm_sda_oe`, releasing SDA. The start condition hold time (t_HD;STA, minimum 600 ns for Fast Mode) requires SDA to stay low while SCL is high for that duration. Releasing SDA after a single system clock cycle will not satisfy this hold time.  We have a state `ENSURE_PLAYBACK_HOLD` that exists specifically to cover this gap but this path bypasses it.

**Possible Fix**: route through `ENSURE_PLAYBACK_HOLD` here too, same as the other branches.

Reproduced in simulation, not caught in hardware by me.

### LINKED ISSUE #499: i2c proxy bug4: SAMPLE_ERROR doesn't reset bit_count
```vhdl
-- line 232-234
when SAMPLE_ERROR =>
    v.state := SAMPLE_IDLE;
```

`bit_count` and `data_bits` are not reset. If the sampler overflows and then re-enters `SAMPLE_START` → `SAMPLE_DATA`, the old `bit_count` is still 7 and the playback comparison in `PLAY_STORED_DATA` will immediately think we've caught up. The `SAMPLE_IDLE` reset case in `SAMPLE_START` clears `bit_count`, but only if there's a new start detected _before_ we re-enter `SAMPLE_DATA`.

**Possible Fix**: reset `bit_count` and `data_bits` in `SAMPLE_ERROR`.

Again by analysis and simulation. I think this is extremely unlikely to actually occur in a real system as we'll see a new start typically but the fix is easy.

nathanaelhuffman:
A number of bugs discovered by analysis, and one found through testing after the other 4 were fixed.

I've got 200 A2 -> A0 sequences on my cosmo with 1.5TB of RAM installed with no missing DIMMs reported, so I think this is good to go.

Fixes #496 
Fixes #497 
Fixes #498 
Fixes #499 

And fixes the bug shown in an image below, which is difficult to reproduce in simulation at the  current simulation bit rates, but was a result of transitioning out of `ENSURE_PLAYBACK_HOLD` if the SP5 SCL was still high.

### DIFF
--- a/hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
+++ b/hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
@@ -6,9 +6,11 @@
 
 -- Unit-level testbench for dimm_arb_mux.
 --
--- Bug 1 cannot be triggered at the integration level: the FPGA abort always finishes
--- before the CPU's first SCL falling edge (~2-3 µs vs the 5 µs STANDARD half-period),
--- so PLAY_STORED_START is entered while sample_r.state=SAMPLE_START and Bug 3 fires
+-- https://github.com/oxidecomputer/quartz/issues/496 cannot be triggered at the integration 
+-- level: the FPGA abort always finishes
+-- before the CPU's first SCL falling edge (~2-3 us vs the 5 us STANDARD half-period),
+-- so PLAY_STORED_START is entered while sample_r.state=SAMPLE_START and 
+-- https://github.com/oxidecomputer/quartz/issues/498 fires
 -- instead.  Direct stimulus control lets us reach PLAY_STORED_DATA with bit_count=1
 -- while keeping cpu_scl_in='1' through the catch-up rising edge.
 
@@ -120,8 +122,8 @@ begin
 
         while test_suite loop
             if run("arb_bug_sda_glitch_during_playback") then
-                -- GH # 469
-                -- Release reset, wait for POWER_UP_CLEAR (~9 *� 1000 ns at FAST_PLUS)
+                -- Issue https://github.com/oxidecomputer/quartz/issues/469
+                -- Release reset, wait for POWER_UP_CLEAR (~9 x 1000 ns at FAST_PLUS)
                 -- and for dimm_i2c_idle to assert (~504 ns with both dimm lines held high).
                 reset <= '1';
                 clk_tick;
@@ -129,7 +131,7 @@ begin
                 wait for 10 us;
 
                 -- Step 1: CPU start condition: SDA falls while SCL is high.
-                -- cpu_start_detected fires� -> mux: IDLE ->�� CPU_REQ_GRANT.
+                -- cpu_start_detected fires -> mux: IDLE -> CPU_REQ_GRANT.
                 -- sample: SAMPLE_IDLE -> SAMPLE_START.
                 cpu_scl_in    <= '1';
                 cpu_sda_in    <= '0';
@@ -139,7 +141,7 @@ begin
                 clk_tick;
 
                 -- Step 2: CPU SCL falling edge.
-                -- sample: SAMPLE_START ->�� SAMPLE_DATA (bit_count still 0).
+                -- sample: SAMPLE_START -> SAMPLE_DATA (bit_count still 0).
                 cpu_scl_in    <= '0';
                 cpu_scl_fedge <= '1';
                 clk_tick;
@@ -155,9 +157,10 @@ begin
                 clk_tick;
 
                 -- Step 4: Grant the bus.
-                -- sample_r.state=SAMPLE_DATA (not SAMPLE_START) �->� GH #498 cannot fire
+                -- sample_r.state=SAMPLE_DATA (not SAMPLE_START) -> 
+                -- Issue https://github.com/oxidecomputer/quartz/issues/498 cannot fire
                 -- even though cpu_scl_in='1'.
-                -- mux: CPU_REQ_GRANT ->�� PLAY_STORED_START (dimm_i2c_idle='1' already).
+                -- mux: CPU_REQ_GRANT -> PLAY_STORED_START (dimm_i2c_idle='1' already).
                 bus_grant <= '1';
                 wait until sp5_playback_i2c_has_bus = '1';
 
@@ -169,13 +172,13 @@ begin
                 -- rising edge, releasing SDA while the playback SCL is still high.
                 --
                 -- Timeline from PLAY_STORED_START entry (playback SCL starts at '1'):
-                --   T + ~500 ns : first playback FEDGE → enter PLAY_STORED_DATA,
+                --   T + ~500 ns : first playback FEDGE -> enter PLAY_STORED_DATA,
                 --                 dimm_sda_oe set to not(data_bits(0))='1' (no change)
                 --   T + ~1000 ns: first playback REDGE, playback_bits=1=bit_count,
-                --                 cpu_scl_in='1' → Bug 1 exits to ENSURE_PLAYBACK_HOLD,
+                --                 cpu_scl_in='1' -> Bug 1 exits to ENSURE_PLAYBACK_HOLD,
                 --                 schedules dimm_sda_oe='0' for next cycle
-                --   T + ~1008 ns: playback_sda_if.oe '1'→'0' while playback_scl_if.o='1'
-                --                 → sda_glitch_monitor fires check_false → test fails
+                --   T + ~1008 ns: playback_sda_if.oe '1'->'0' while playback_scl_if.o='1'
+                --                 -> sda_glitch_monitor fires check_false -> test fails
                 cpu_sda_in <= '1';
                 wait for 2 us;
 
@@ -183,9 +186,9 @@ begin
                 wait for 500 ns;
 
             elsif run("arb_bug_playback_stall") then
-                -- Direct stimulus reproduction of Bug 2.
+                -- Direct stimulus reproduction of https://github.com/oxidecomputer/quartz/issues/497.
                 --
-                -- Bug GH # 497: in PLAY_STORED_DATA, when playback_bits catches up to bit_count
+                -- Bug: in PLAY_STORED_DATA, when playback_bits catches up to bit_count
                 -- at a falling playback SCL edge while cpu_scl_in='1', the exit condition
                 -- `playback_scl_fedge='1' and playback_bits=bit_count and cpu_scl_in='0'`
                 -- is FALSE because cpu_scl_in is '1'.  The following REDGE overshoots to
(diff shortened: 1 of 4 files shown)


hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
```diff
@@ -0,0 +1,249 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at https://mozilla.org/MPL/2.0/.
+--
+-- Copyright 2025 Oxide Computer Company
```
Aaron-Hartwig, line 5:
```suggestion
-- Copyright 2026 Oxide Computer Company
```

Do we need this statement even? Over in hubris we don't have it and just use those first 3 MPL lines.

nathanaelhuffman:
With changes at this point, there's at least one bug here still, this looks like it might be around playback sclk fedges?:
image

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
nathanaelhuffman: We do not, this shows as optional in the open source policy. I have removed all of these lines from this block. I may consider doing another PR that does this change throughout the whole repo clean we'll see.

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
```diff
@@ -0,0 +1,252 @@
...
+    end process;
+
+    bench: process
+        procedure clk_tick is begin wait until rising_edge(clk); end procedure;
+    begin
+        test_runner_setup(runner, runner_cfg);
+
+        while test_suite loop
+            if run("arb_bug_sda_glitch_during_playback") then
+                -- GH # 469
```
Aaron-Hartwig:
```suggestion
                -- GH # 496
```
or lets just slap the direct link to the issue here?

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
```diff
@@ -0,0 +1,252 @@
...
+                -- Release reset, wait for POWER_UP_CLEAR (~9 *� 1000 ns at FAST_PLUS)
```
Aaron-Hartwig: There are a lot of � in this source, those should get cleaned up

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
```diff
@@ -0,0 +1,252 @@
...
+                --
+                -- data_bits(0)='0'  playback currently drives SDA low (oe='1').
+                -- Bug 1 will set oe := not(cpu_sda_in) = not('1') = '0' on the catch-up
+                -- rising edge, releasing SDA while the playback SCL is still high.
+                --
+                -- Timeline from PLAY_STORED_START entry (playback SCL starts at '1'):
+                --   T + ~500 ns : first playback FEDGE → enter PLAY_STORED_DATA,
+                --                 dimm_sda_oe set to not(data_bits(0))='1' (no change)
+                --   T + ~1000 ns: first playback REDGE, playback_bits=1=bit_count,
+                --                 cpu_scl_in='1' → Bug 1 exits to ENSURE_PLAYBACK_HOLD,
```
Aaron-Hartwig: Having `→` in places still reads very "An LLM wrote this" to me as I've never really seen humans use that over `->`.

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
```diff
@@ -0,0 +1,252 @@
...
+                --   T + ~1008 ns: playback_sda_if.oe '1'→'0' while playback_scl_if.o='1'
+                --                 → sda_glitch_monitor fires check_false → test fails
+                cpu_sda_in <= '1';
+                wait for 2 us;
+
+                bus_grant <= '0';
+                wait for 500 ns;
+
+            elsif run("arb_bug_playback_stall") then
+                -- Direct stimulus reproduction of Bug GH # 497.
```
Aaron-Hartwig: Again, maybe just link right to the bug?

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
```diff
@@ -0,0 +1,252 @@
...
+                -- Release cpu_scl_in to allow ENSURE_PLAYBACK_HOLD to proceed to CPU_HAS_BUS.
+                cpu_scl_in <= '0';
+                wait until sp5_i2c_has_bus = '1' for 10 us;
+                check_true(sp5_i2c_has_bus = '1',
+                    "Bug GH # 497: mux stalled in PLAY_STORED_DATA -- never reached CPU_HAS_BUS");
+
+                bus_grant <= '0';
+                wait for 500 ns;
+            end if;
+        end loop;
```
Aaron-Hartwig, line 330: Did we lose the test cases for 498 and 499? I'm just curious why we have test cases for all four at the SPD proxy top level but then just the first two at the arb mux level.

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
nathanaelhuffman: fixed

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
nathanaelhuffman: fixed

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
nathanaelhuffman: fixed. unicode characters also aren't allowed in VHDL by spec :D

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
nathanaelhuffman: Fixed

hdl/projects/cosmo_seq/dimms_subsystem/sims/dimm_arb_mux_tb.vhd
nathanaelhuffman: I moved stuff around and there was a bunch of copy paste crap left over. we now have 1 test in the top level and 3 tests in the arb_mux_tb as it's easier to exhibit the errors and test the fixes there.

PR #502, opened 2026-05-15
fpga-releaser: fix --skip-gh and inverted needs_gh check

nathanaelhuffman:
#500 broke the normal release flow due to a logic bug that makes a normal run with no --skip-gh and no --zip fail:

The change in #500 is equivalent to `not args.skip_gh and args.zip is not None`, so for the default invocation (no --skip-gh, no --zip) needs_gh evaluated to False, meaning the `api` stayed None, and `get_latest_artifact_info` blew up dereferencing that None.

In the original intention of this code, skip-gh really meant skip-gh-*release*.  Even when fixing the bug above, the system still accessed the github api when skip-gh was set so this PR actually fixes that problem also so skip-gh really does skip github now, releases and APIs.

### DIFF
--- a/tools/fpga_releaser/cli.py
+++ b/tools/fpga_releaser/cli.py
@@ -48,8 +48,14 @@ def main():
     if args.zip:
         project_info.local = True
 
-    # Only require a GitHub token when we actually need to talk to GitHub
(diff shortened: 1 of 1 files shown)


Aaron-Hartwig (approved):
oops :P

PR #508, opened 2026-07-17
rdl: fix improperly configured field sw access

### LINKED ISSUE #507: `sw_access` in JSON output doesn't seem to match HDL
For `post_code_monitor`, the fields are marked read only to software:

https://github.com/oxidecomputer/quartz/blob/6f68cc720f900e2fddd1cb156ef4cd99bcad4238/hdl/ip/vhd/espi/sys_regs/espi_regs.rdl#L131-L135

But in the generated JSON, the sw_access [is marked as](https://github.com/oxidecomputer/hubris/blob/53a149bc5f3451b1e1ce030a7a0043012d0d57e2/drv/spartan7-loader/cosmo-seq/espi_regs.json#L346-L356) "rw":

```json
{
    "type": "field",
    "inst_name": "phbl_hello",
    "lsb": 7,
    "msb": 7,
    "reset": 0,
    "sw_access": "rw",
    "se_onread": null,
    "se_onwrite": null,
    "desc": "0x1de90001 - PHBLhello. Sticky: set when post code is seen, cleared by hardware on new boot."
}
```

This leads to the codegen emitting setters for these fields, which is perhaps not what we want!

Aaron-Hartwig:
In SystemRDL, the `default` keyword only applies to children within the scope it is declared. This means when it is used within a `field` (which can't have children), nothing happens. Additionally, it does not set the property locally (that is, on the `field`), so all the places where `default $property = $value` exist within a field don't actually do anything. This PR fixes all the occurrences where we did this within our RDL definitions.

Fixes #507

### DIFF
--- a/hdl/ip/vhd/espi/sys_regs/espi_regs.rdl
+++ b/hdl/ip/vhd/espi/sys_regs/espi_regs.rdl
@@ -21,8 +21,8 @@ addrmap espi_regs {
         desc = "";
 
         field {
-            default sw = woclr;
             desc = "Sticky bit for alert, set to 1 when alert is needed, cleared by writing 1";
+            woclr = true;
         } alert[0:0] =  0;
     } flags;
 
@@ -49,9 +49,9 @@ addrmap espi_regs {
     reg {
         name = "Status Register";
         desc = "";
+        default sw = r;
 
         field {
-            default sw = r;
             desc = "Set to one to 1 when hw is running the spi transaction, no new transactions
             may be issued until it is finished. Technically represents cs_n being low.";
         } busy[0:0] =  0;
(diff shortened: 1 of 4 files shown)


hdl/ip/vhd/espi/sys_regs/espi_regs.rdl
```diff
@@ -21,8 +21,8 @@ addrmap espi_regs {
         desc = "";
 
         field {
-            default sw = woclr;
```
Aaron-Hartwig, line 24: This is actually an invalid statement which was swallowed by the compiler since `default` is meaningless within `field`. `woclr` is a bool type that can be set on its own, so I adjusted that.

jamesmunns:
Great catch! Do we need to do anything to update the JSON files checked into `hubris` once this change has landed?

nathanaelhuffman:
> Great catch! Do we need to do anything to update the JSON files checked into `hubris` once this change has landed?

Yep we'll need to push new releases to hubris with the json files I'll work on making that happen this week.

nathanaelhuffman (approved):
ty for digging into this

PR #513, opened 2026-08-05
rewrite the axi fabric to support pipelines.

### LINKED ISSUE #480: We're starting to feel timing pressure on the axi interconnect as the design scales
We should probably consider pipe-lining each of the de-muxed interfaces at the very least. I'd like to make this a bit more fully featured longer term.

nathanaelhuffman:
As #480 mentions, our somewhat naive axi fabric has some improvement opportunities.

This PR rewrites the bock to support user-specified pipeline stages between each of the axi responders, and adds more test coverage since the block is not as simple. Some additional logic improvements are integrated in the rewrite including a one-hot decode methodolgy to reduce logic and increase fmax.

This also fixes a couple of bugs in the axi8 responder that we weren't experiencing in hardware given the previous design but would experience now given the implementation removed some of the simplifications:
- responder read address was driven from the write address
- decode teardown was unreachable while AWVALID stayed asserted, so the next read decoded against a stale address
-  a request held one cycle past its handshake re-armed the fabric and issued a duplicate transaction
- and axil8_resizer had four reversed assignments and reads were't simulated so this was not caught in simulation.

This is running successfully on cosmo hw.

Fixes  #480 

Summary of new functionality:
- Decode/mux rework: range membership became an equality compare on the address bits above each responder's span (bases are span-aligned, spans are powers of two), replacing two 32-bit magnitude compares per responder and their carry chains. 
- The dynamic address mask which was implemented as a variable-shift mux shared by every responder path is now  compile-time constant per responder.  The responder select became a registered one-hot instead of an integer index, so the return path is a flat AND-OR tree.
- Configurable pipelining: pipe_stages per responder in the config record. Since the fabric admits one transaction at a time, axil_pipe serializes it into one request bundle out and one response bundle back rather than five channel register slices.  This new implementation uses about a third of the flops, one 1-bit token chain, no timing exceptions required (everything is register-register timed). If you set pipe_stages = 0, you get a free pass-through.

A fun timing comparison:

| image | before |after |
|--- | --- | --- |
| cosmo_seq | WNS 0.100 ns | 0.120 ns (worst path moved out of eSPI) |
| grapefruit | WNS 0.038 ns | 0.226 ns |
| cosmo_hp | 86.2 MHz | 101.5 MHz |

Note that most of this gain is from the consolidated logic: only cosmo_seq's eSPI responder is piped, at one stage.


hdl/ip/vhd/axi_blocks/axil_pipe.vhd
```diff
@@ -0,0 +1,383 @@
...
+-- its way back.
+--
+-- Round trip latency is 2 * stages cycles plus a handful of fixed handshake
+-- cycles. stages = 0 is a plain pass-through and costs nothing.
+
+entity axil_pipe is
+    generic (
+        --! Register stages inserted in *each* direction
+        stages : natural;
+        --! Address bits this responder actually decodes
```
Aaron-Hartwig: I don't think we need doc comments formatted with `--!` anymore, just use `--`.

hdl/ip/vhd/axi_blocks/axil_pipe.vhd
```diff
@@ -0,0 +1,383 @@
...
+
+        signal fwd_tok : std_logic_vector(stages - 1 downto 0);
+        signal ret_tok : std_logic_vector(stages - 1 downto 0);
+
+        type sink_state_t is (armed, wait_rsp, respond, request_gap);
+
+        signal sink_state : sink_state_t;
+        signal sink_is_write : std_logic;
+        --! registered one-shot, so AW and W always handshake in the same cycle
+        --! and AWREADY never depends combinationally on AWVALID
```
Aaron-Hartwig: more `--!`

hdl/ip/vhd/axi_blocks/sims/axil_interconnect_sim_pkg.vhd
```diff
@@ -0,0 +1,70 @@
...
+         WIDE_IDX   => resp_cfg(base_addr => x"00008000", addr_span_bits => 15, pipe_stages => 2));
+
+    --! An integer as a bus address
+    function ba (constant addr : integer) return std_logic_vector;
+
+    --! Base address of a responder, plus a byte offset, as a bus address
+    function ba (constant idx : integer; constant offset : integer) return std_logic_vector;
+
+    --! An integer as a 32 bit data word
+    function w32 (constant value : integer) return std_logic_vector;
```
Aaron-Hartwig: More `--!` materializing. And as a point of personal taste, I think this naming is rather laconic.

hdl/ip/vhd/axi_blocks/sims/axil_slow_responder.vhd
```diff
@@ -0,0 +1,203 @@
...
+        rready : in    std_logic;
+        rdata  : out   std_logic_vector(31 downto 0);
+        rresp  : out   std_logic_vector(1 downto 0)
+    );
+end entity;
+
+architecture rtl of axil_slow_responder is
+
+    constant NUM_WORDS : integer := 16;
+    constant MAX_DELAY : integer := 3;
```
Aaron-Hartwig: I think we should consider making `MAX_DELAY` a generic port on this module, and then set that (and the corresponding `responder.pipe_states` in the pkg) via the same constant that lives in `axil_interconnect_sim_pkg.vhd`. Just so there's one constant driving consistent behavior here.

Aaron-Hartwig (approved):
A few nits but this LGTM. I know you've had a bit of runtime through other testing as well.

hdl/ip/vhd/axi_blocks/sims/axil_slow_responder.vhd
nathanaelhuffman: done

Aaron-Hartwig (approved):

PR #514, opened 2026-08-06
Ndh/speed up spi nor

nathanaelhuffman:
I've stacked this for now on the axi-pipes branch, but will rebase this against main once axi-pipes lands. This allows review of just this change.

Here we increase the SPI nor frequency to 62.5MHz (up from ~20MHz). This required some link layer optimization and the addition of some timing constraints for this.  Simulation coverage has also increased.

At least when benchmark over usb, this basically make 0 difference as the usb transfers dominate the performance there.

Given what we know about flash and the system performance, it is unclear whether this will speed anything up in product but it gets the FPGA shift speed out of the way for any future performance investigation.  There is probably additional optimization to be had around queueing up more data or other interface changes here should any hubris benchmarking suggest additional optimization opportunity.

Change summary:
Speed up sclk 20.83 MHz → 62.5 MHz (clk/6 → clk/2 off clk_125m) . 
The register map untouched, so hubris code requires no changes. 
Sclk rate and sample point are new generics on spi_nor_top.

A number of bugs that were not evident at the slower clock frequency were fixed as they were evident at faster speeds:
- opcode serialized in the transaction's data mode during cs_assert. This would result in an incorrect opcode sent to the device.
- OE followed the byte-aligned mode, resulting in IO3 re-driven into the part on a quad read turnaround.
- output teardown at end of write, resulting in corrupting the last sample edge at the flash part.
- `sclk_running` conflated with `in_tx_phases` resulted in trailing clocks during cs_deassert. Erase and write-enable could be silently discarded in this case.
- io mode taken from the registered state at the reload edge resulted in the first quad data byte shifted by 1 not 4, so we lost the sentinel in the register and the transaction never terminated.

**Structural Changes:** 
TX launches on the sclk falling edge itself; byte and io mode are prefetched together; RX sampling moved to a placed point (rx_sample_taps, half-clk steps) fed by rising/falling capture flops; every pin-facing flop duplicated with no internal fanout and packed into its IOB to close timing.  (IOB packing reduces 12–13 ns of routing  to ~3.3 ns).

This has been tested in hardware on a cosmo with the following somewhat adhoc tests via humility:

```
host flash slot: Flash0
test region:     0x400000 .. 0x410000

[id] read JEDEC ID + unique ID
    mfr=0xef type=0x40 capacity=0x21
    uid=ea626c4097221b35ea626c710313313600
    PASS (1.5s)

[status] read status register, check WIP clear
    status = 0x00
    PASS (1.4s)

[erase] sector erase leaves every byte of the sector 0xFF
    PASS (5.7s)

[page_program] program a page and read it back
    PASS (5.5s)

[partial_page] sub-page write of a non-multiple-of-4 length
    PASS (5.7s)

[multi_page] program four consecutive pages, verify the whole sector
    PASS (8.3s)

[odd_lengths] reads of awkward lengths at awkward offsets
    8 length/offset combinations verified (last one crosses a page boundary)
    PASS (15.2s)

[bit_clear] reprogramming without erase only clears bits
    PASS (7.3s)

[ff_write_noop] writing all-0xFF is a documented no-op
    PASS (5.8s)

[erase_after_write] erase clears a programmed sector, and is idempotent
    PASS (7.4s)

[neighbor_isolation] operations do not disturb the adjacent sector
    PASS (5.5s)

[diffwrite] -D programs a full sector; a second -D finds no delta
    second pass reported no delta
    PASS (12.9s)

[bad_address] out-of-slot access is rejected, not silently wrapped
    rejected as expected
    PASS (1.4s)

[sector0_guard] sector 0 is refused without --write-sector0
    refused as expected
    PASS (1.1s)

[throughput] time a full-sector -D write and a full-sector hash pass
    hash (read) pass: 1.46s -> 43.9 KiB/s including invocation overhead
    write pass:       7.40s -> 8.7 KiB/s (erase + program)
    PASS (10.5s)
```


hdl/projects/cosmo_seq/cosmo_timing.xdc
```diff
@@ -171,4 +171,116 @@ set_false_path -from [get_ports {*}] -to [get_ports {fpga1_spare_v1p8[*]}]
...
+# sclk is toggled by fabric logic off clk_125m at clk/2, so 62.5MHz, a 16ns
+# period with an 8ns half period. Nothing inside the FPGA is clocked by it, so
+# there is deliberately no create_generated_clock here: what actually has to be
+# bounded is the clock-to-data skew leaving the FPGA and the pin-to-flop delay
+# coming back, and both are directly constrainable.
+#
+# Trace delays are short and local; using the same 6.8ns/m as the FMC block
+# above. TODO: replace with the measured lengths off 913-0000023.
+set flash_trace_max 0.40
+set flash_trace_min 0.10
```
Aaron-Hartwig: Do we want to get real values here before merging? The comment below about input timing constraints on the data pins would also need to be adjusted.

hdl/ip/vhd/spi_nor_controller/sims/spi_nor_tb_pkg.vhd
```diff
@@ -43,6 +45,87 @@ package spi_nor_tb_pkg is
...
+    -- or a neighbouring bit, and lands here as a mismatch.
+    procedure check_read_pattern (
+        signal net      : inout network_t;
+        constant addr   : natural;
+        constant count  : natural
+    );
+
+    -- Checked scenarios, shared by the testbenches so the same coverage can be
+    -- run at more than one sclk rate without duplicating it.
+
```
Aaron-Hartwig: extra space here

hdl/ip/vhd/spi_nor_controller/sims/spi_nor_tb.vhd
```diff
@@ -11,6 +11,7 @@ library vunit_lib;
     context vunit_lib.vunit_context;
 use work.spi_nor_tb_pkg.all;
 use work.spi_nor_pkg.all;
+use work.spi_nor_target_vc_pkg.all;
```
Aaron-Hartwig, line 14: I can't comment on unchanged lines, but the `spi_nor_th` instantiated down on L26 below now has a bunch of generics defined, so this TB is relying on their defaults. Maybe that is fine since then we have a TB that always is testing the defaults and there are other TBs for specific configurations, but I wanted to highlight it anyway.

hdl/projects/cosmo_seq/sp5_espi_flash_subsystem/sp5_espi_flash_subsystem.vhd
```diff
@@ -152,6 +152,21 @@ begin
 
 
     spi_nor_top_inst: entity work.spi_nor_top
+        generic map(
```
Aaron-Hartwig, line 155: Given this is a design-specific instantiation, I think we should exhaustively set the generics, even if they're currently the same as the default of the IP block.

hdl/ip/vhd/spi_nor_controller/sims/spi_nor_fast_quick_io_tb.vhd
```diff
@@ -0,0 +1,95 @@
...
+begin
+
+    th: entity work.spi_nor_th
+        generic map (
+            sclk_divisor   => 0,
+            rx_sample_taps => 2,
+            -- Fast corner of the delays the XDC bounds
+            out_delay => 1.6 ns,
+            in_delay  => 0.5 ns
+        );
```
Aaron-Hartwig, line 47: No action, just a note. It would be nice if we could run the same TB against THs configured differently. Either teach VUnit how to do that, do some sort of pre-processing and then point VUnit to a set of generated TBs which were identical aside from the TH.

Aaron-Hartwig (approved):
Some small notes you can take or leave.

hdl/projects/cosmo_seq/cosmo_timing.xdc
nathanaelhuffman: yep, though real values are even better than this. fixed in review cleanup commit inbound.

hdl/ip/vhd/spi_nor_controller/sims/spi_nor_tb.vhd
nathanaelhuffman: made them explicit

hdl/projects/cosmo_seq/sp5_espi_flash_subsystem/sp5_espi_flash_subsystem.vhd
nathanaelhuffman: Fixed.

hdl/ip/vhd/spi_nor_controller/sims/spi_nor_fast_quick_io_tb.vhd
nathanaelhuffman: This is possible in vunit, but how this works with our generated stuff is ?? I'm going to open an issue so we don't lose this idea but punt on it for this change.

PR #516, opened 2026-08-10
Add sha3-256 hardware for hasing offload of host image.

### LINKED ISSUE #510: FPGA assisted host flash SHA calcuation
As part of the work for RFD 682  Measuring & attesting to HBS we need to calculate the sha3-256 hash of the host image that's booted. We already do this on the SP for other system requirements. The SP reads out the data from QSPI and uses a hardware block to calculate the hash. This takes about 18 seconds as of this writing. There are probably optimizations we can do on the SP side but it would be a significant improvement if there
were an interface to have the FPGA calculate this hash.

Requirements/questions:
- I think this should actually be faster than having the SP do the work (no crypto decelerators here) but plausibility check is important
- We don't want to include sector0 in the hash calculations but need to treat it as erased/all `0xff`
- Having this calculation happen automatically without SP intervention would be nice but might be unrealistic
- sha3-256 is our current requirement as of this writing. We should double check how soon we expect to switch to something else.

A register interface might look like:
- Write to start a new hash
- Register to allow SP to feed in data to hash block
- Register to give start address and len to read QSPI data into the hash
- Register to calculate the final resut
- Status/error register for polling/waiting

nathanaelhuffman:
As describe in #510, this provides a sha3-256 hash offload block that fetches data from spi-nor based on the requested parameters (size, starting location, how many bytes of 0xFF to pre-fill to cover sector0 etc) and returns the sha3-256 hash.

This has been tested on cosmo hardware and against sw sha3-256 implementations and produces matching hashes.

Fixes #510


tools/hash_engine_flash_test.py
```diff
@@ -0,0 +1,561 @@
...
+ST_ABORTED = 1 << 4
+ST_CFG_ERR = 1 << 5
+
+# CONFIG.source encoding.
+SRC_LOCAL_REG = 0
+SRC_HOST_QSPI = 1
+
+# 4 KiB, matching SECTOR_BYTES in the SPI NOR verification component. Override
+# with --sector-size if the part in question uses something else.
+DEFAULT_SECTOR_SIZE = 0x1000
```
labbott: The default sector size in hubris is `0x1_0000` . Not a huge deal but might make it more representative

Aaron-Hartwig (approved):

PR #522, opened 2026-08-19
sidecar: use RDL-defined types in sequencer

### LINKED ISSUE #521: sidecar: make use of RDL-defined types in  `Tofino2Sequencer.bsv`
We're defining them in the RTL which is fragile since we can't rely on codegen to pickup any encoding updates. This should be addressed to use the RDL encoded types so the BSV and Rust can all use the same stuff.

This is blocking https://github.com/oxidecomputer/hubris/pull/2515/

Aaron-Hartwig:
This moves the state, step, and error types for the Tofino sequencer code out of RTL and into the RDL definition. There are no functional changes here, just shuffling types around. This will support allowing the Hubris sidecar-seq code to leverage the same types as the FPGA. Currently we duplicate the type definitions, once in the BSV and again in Rust.

Fixes #521

### DIFF
--- a/hdl/projects/sidecar/mainboard/PCIeEndpointController.bsv
+++ b/hdl/projects/sidecar/mainboard/PCIeEndpointController.bsv
@@ -87,7 +87,7 @@ module mkPCIeEndpointController
         let sequencer_power_fault =
                 ctrl.present == 1 &&
                 ctrl.override_seq_power_fault == 0 &&
-                sequencer.registers.error.error != 0;
+                unpack(sequencer.registers.error.value) != None;
 
         power_fault <= (software_power_fault || sequencer_power_fault);
     endrule
--- a/hdl/projects/sidecar/mainboard/SidecarMainboardController.bsv
+++ b/hdl/projects/sidecar/mainboard/SidecarMainboardController.bsv
@@ -194,11 +194,15 @@ module mkMainboardController #(Parameters parameters)
 
     (* fire_when_enabled *)
     rule do_set_status;
+        // Binding the unpacked value to a typed signal to disambiguate `Init`,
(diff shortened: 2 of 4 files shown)


hdl/projects/sidecar/mainboard/PCIeEndpointController.bsv
```diff
@@ -87,7 +87,7 @@ module mkPCIeEndpointController
         let sequencer_power_fault =
                 ctrl.present == 1 &&
                 ctrl.override_seq_power_fault == 0 &&
-                sequencer.registers.error.error != 0;
+                unpack(sequencer.registers.error.value) != None;
```
nathanaelhuffman, line 90: Will this work when we go to rust as "None" has special semantics there?

nathanaelhuffman (approved):
One tiny comment about whether None becomes overloaded or confusing once this is used in Rust, otherwise LGTM

hdl/projects/sidecar/mainboard/PCIeEndpointController.bsv
Aaron-Hartwig: Yeah, at least it is happily building over on https://github.com/oxidecomputer/hubris/pull/2652. Everything in BSV is scoped to the package, so there is not really a notion of `Type::Member` since it basically just becomes `Package::Member`. This is why I had to do the shenanigans in `SidecarMainboardController.bsv` to give `bsc` a way to infer _which_ `Init` was being compared, because from `bsc`'s perspective there were multiple definitions of `SidecarMainboardControllerRegs::Init` since I can't do something like `== TofinoSeqStateValue::Init` or `== TofinoSeqStepValue::Init`. In Rust these members are properly scoped within the type, so we'll have `TofinoSeqErrorValueEncoded::None`, not just `None` like in the BSV).

PR #524, opened 2026-08-21
rdl: a number of changes to how generated register maps are handled

Aaron-Hartwig:
A number of commits here:
1) With the migration to buck2, adoc generation was dropped. This change adds it back and updates the generation to account for the way flat register maps are now generated. Additionally, we had a bug in the HTML template where we'd generate an empty table row for every single byte of a memory. In practice this wouldn't get rendered, but it would massively inflate the size of the HTML document itself (on cosmo_seq it goes from 4MB down to 520KB). 
2) Have cosmo_seq output an adoc regmap
3) We'll also check that BSV projects have reg maps added to their deps so those intermediate artifacts (like reg maps) end up in the build artifacts.
4) Cleanup how regmaps are handled (no longer be `hidden`)
5) Ouput adoc for BSV project which currently don't.


nathanaelhuffman (approved):
ty!

PR #527, opened 2026-09-10
fmc: streamline the FMC target and run it at 100 MHz on cosmo

nathanaelhuffman:
This contains a bunch of work to speed up the FMC interface to our theoretical limits here.  Hubris currently runs the FMC inteface at 50MHz and the measured data rates are around 8.3MBps at this speed.  We've re-tooled both the FMC state machine, and the timing constraints to allow this to get up to 100MHz (on cosmo/metro) and 66MHz (on grapefruit due to SP <-> FPGA pinout errata which limits our max freq due to not having fmc_clk on a clock-capable pin).

We hit 100MHz by phase shifting an internal representation of the fmc_clk forwards and backwards, depending on input/output direction to cancel out these delays as well as using IOB packing. Discussion in detail of this is in the documentation.

This change is backwards compatible with hubris such that it works with the existing clock (50MHz) as well as 66 or 100MHz without needing RTL changes. This is done by choosing PLL parameters for the shifted clocks that are functional between 50 and 100MHz and doing the timing constraints at 100MHz.

Benchmarking:
50MHz FMC bus: ~8.3MBps
66MHz FMC bus: ~9.9MBps, 
100MHz FMC bus: ~11.4 MBps.

For validation:
I've run multiple hours of read/write hammering using the fmc_demo UDP blocks, including new features to be upstreamed in hubris that run the loops as fast as possible without network being the load-bearing factor.

The phase of the re-clocked FMC pins  was phase shifted to prove that we're in the middle of the sample window and run both the + and - shifted images for 1 hour each with constant hammering of the FMC block which validates that we have margin on our sample points even though timing appears tight (but passing!). It is expected that this timing will remain tight, but not have much build-build variance due to it being limited more by physics than synth and placement.


hdl/ip/vhd/fmc_if/docs/fmc_verification_plan.adoc
```diff
@@ -0,0 +1,203 @@
+:showtitle:
```
Aaron-Hartwig, line 1: This probably doesn't have to be checked in, but I don't think it harms anything if it is.

Aaron-Hartwig (approved):

